PR #2, opened 2023-10-03
towards a working spawn

nick1udwig:


dr-frmr:
spawn "works" now? merge time

barraguda:
so in [fb7e2a9](https://github.com/uqbar-dao/uqbar/pull/2/commits/fb7e2a924a6a3b69c9797db8ac3458ec731e8541) I took  some fixes made to vfs in #13 , and it's almost there.

there is a central problem here we need to address, which is how vfs interacts with filesystem. we don't really have awaits between runtime modules currently, before this commit vfs wouldn't boot up as it would wait for it's getstate function forever, because only way to await a message is to recv() in the main receiver loop.

I think we could setup an extra mspc channel (2) between fs and vfs, so vfs can await calls without blocking the main loop. I'll try this out next.

src/vfs.rs
```diff
@@ -740,20 +724,32 @@ async fn match_request(
                         .await;
                     let write_response = recv_response.recv().await.unwrap();
```
barraguda, line 768: ^essentially anywhere we do this it's a problem

dr-frmr: yes. extremely

nick1udwig:
Status update:

As of 9f1d0a7

What is working:

* bootstrap fr fr
* Caps set up properly
* vfs set up properly
* vfs now has queues per drive
* spawn close to working properly -- there is a StartedProcess that doesn't get into it properly for some reason
* key_value can spin up a _worker

What is not working:

* Login to FE does not work
* spawn weirdness w/ StartedProcess as mentioned above
* key_value & _worker caps not quite right
* _worker redb integration is not yet tested -- assuming there will be bugs

@bitful-pannul @dr-frmr

nick1udwig:
https://github.com/uqbar-dao/uqbar/blob/b888191fb138ac4d1960544104402a3a12fe6f4e/src/kernel/mod.rs#L1645-L1685

We should consider removing this: kernel modules can directly send on the caps sender (which allows them to await the Response -- important for proper ordering). This would require us to add another import for processes, though; some equivalent of grant_capability() where issuer is set by kernel and params are set by caller.

@dr-frmr @bitful-pannul 

EDIT: It will currently deadlock due to the rx.await. Commenting that out would make it functional, but the Response it sends would be useless since we have no guarantee the cap has been added by the time the Response fires.

PR #6, opened 2023-10-03
RPC: start a process

willbach:


dr-frmr (approved):
I think this is good. Nick and I worked today on a general strategy inside the kernel for launching new processes, and it will have somewhat of an impact on this, but this works for now and contains important upgrades to kernel.

PR #9, opened 2023-10-04
fs errors refactor

barraguda:
Doesn't change how processes currently switch on Result<FsResponse, FsError>

Tried to go for a flow where you'll get a general (this action failed: {action}, with this inner detailed error: {error})

note: we need to clean up vfs panics [x], ~~and u128 -> u64 for file_uuids~~

### DIFF
--- a/src/filesystem/mod.rs
+++ b/src/filesystem/mod.rs
@@ -506,7 +506,7 @@ async fn handle_request(
             };
 
             let file = FileIdentifier::Process(source.process.clone());
-            match manifest.append(&file, &payload.bytes).await {
+            match manifest.write(&file, &payload.bytes).await {
                 Ok(_) => (),
                 Err(e) => {
                     return Err(FsError::WriteFailed {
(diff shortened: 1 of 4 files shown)


src/filesystem/mod.rs
```diff
@@ -469,16 +506,30 @@ async fn handle_request(
...
-                return Err(FileSystemError::BadBytes {
-                    action: "Write".into(),
+                return Err(FsError::BadBytes {
+                    action: "SetState".into(),
                 });
             };
 
             let file = FileIdentifier::Process(source.process.clone());
-            let _ = manifest.write(&file, &payload.bytes).await;
+            match manifest.append(&file, &payload.bytes).await {
```
nick1udwig: I noticed this changed from write to append. Is this intended?

nick1udwig (approved):
Haven't tested, LGTM

src/filesystem/mod.rs
barraguda: very good catch!

barraguda: oh nvm this should be .append, thought I had put it in the wrong place

PR #10, opened 2023-10-05
frontend enforces 9 character minimum

tadad:
See this commit:
https://github.com/uqbar-dao/register/commit/9c0d0334b39ec07c4a3bfb7d5ce5b7a5f6cf7a76


dr-frmr:
@tadad assuming we should close this and prefer #18 ?

tadad:
yes - though I'm not sure why this one wasn't merged...hard to introspect when it's not a submodule which is fixed in #18

PR #11, opened 2023-10-06
docker

barraguda:
WIP

to run, build modules folder, and run following docker command with home_dir reference:

```
docker build -t releaseuqbar .
 
docker run -it -p 8080:8080 --name uq1 \
-v $(pwd)/your_home_directory:/app/userdir \
-v $(pwd)/modules:/app/modules \
releaseuqbar /app/userdir --rpc your_rpc_endpoint
```

todos: 
- lock down cargo version + crossplatform images [x]
- mem usage currently ~50-70mb, figure out with heaptrack/cargo instruments how much of this is wasmtime etc [x]
- nodejs program that calls out to docker to start/stop nodes, controller app in node that calls this via http []

~~also better /modules folder, for new nodes to tap into while bootstrapping~~ , + make terminal use internal fs?


barraguda:
given that we have binary releases now with the core modules built in, a docker deploy would just need to use that binary, and link a local folder for the user directory!

PR #24, opened 2023-10-14
key_value fr

nick1udwig:
Working great as of 343a2ce.

I want to figure out how to make importing deps foolproof, then will be ready to merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Last updated: 10/02/23
 # Clone the repo.
 
 git clone git@github.com:uqbar-dao/uqbar.git
+git clone git@github.com:uqbar-dao/redb.git
 
 # Get some stuff so we can build wasm.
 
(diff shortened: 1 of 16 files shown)


nick1udwig:
@dr-frmr this PR is ready to merge with changes to README.

Users will have to get redb repo, as in README.

Currently redb repo is private, so people who want to build uqbar will have to have access to both redb & uqbar, or we can make redb public (there is nothing secret there, and it can serve as an example for importing libs -- though some error handling work still needs to be done there).

Current users will also have to be told to re-follow the README to make sure they have redb.

Merge whenever ready @dr-frmr

nick1udwig:
Usage example:
```
!message our key_value:key_value:uqbar {"New": {"db": "omegalul"}}
!message our key_value:key_value:uqbar {"Write": {"db": "omegalul", "key": [1, 2, 3], "val": [4, 5, 6]}}
!message our key_value:key_value:uqbar {"Read": {"db": "omegalul", "key": [1, 2, 3]}}
```
the `Read` will print if you have verbose mode on.

Works for persistence as well -- you need to re-initialize the db with a `New`, but then you can `Read` your previously written data.

PR #27, opened 2023-10-16
fs: unify Write and Replace(id)

barraguda:
leads to better cleanup of old files written/overwritten by vfs, also better interface.


src/vfs.rs
```diff
@@ -806,7 +806,7 @@ async fn match_request(
                             message: Message::Request(Request {
                                 inherit: true,
                                 expects_response: Some(5), // TODO evaluate
-                                ipc: Some(serde_json::to_string(&FsAction::Write).unwrap()),
+                                ipc: Some(serde_json::to_string(&FsAction::Write(None)).unwrap()),
```
barraguda: changed these to write(none) for now to compile, but we should probably maybe check for existing here? @hosted-fornet

nick1udwig: Yeah, I think you are right. Do you understand this code well enough to make this change or should I push a commit here/elsewhere?

barraguda: if you have time to push a commit here, would be great! @hosted-fornet

nick1udwig: Shoot, forgot about this @bitful-pannul my bad

nick1udwig: Added 9014ff3 @bitful-pannul

src/vfs.rs
```diff
@@ -782,22 +782,30 @@ async fn match_request(
...
+                            match vfs.path_to_key.remove(&full_path) {
+                                None => None,
+                                Some(key) => {
+                                    let Key::File { id: hash } = key else {
+                                        panic!("");
+                                    };
+                                    Some(hash)
+                                }
+                            }
+                            // vfs.key_to_entry.remove(&old_key);
```
barraguda, line 806: @hosted-fornet added the same thing for files within a zip directory too, but what's this key_to_entry removal? in the case of an old file lingering at path?

nick1udwig: Take a look at the data structures and think about it: https://github.com/uqbar-dao/uqbar/blob/82b6baa573067127685c4181683d8fa33bd0f864/src/vfs.rs#L15-L21

Ask me again if you still can't figure it out. This is a pretty important architectural feature of the vfs (hint as to why: https://github.com/uqbar-dao/uqbar/blob/82b6baa573067127685c4181683d8fa33bd0f864/src/vfs.rs#L43-L44)

barraguda: oh yes I see, but it's not needed here as it's the same key to entry as before right, so unnecessary print?

PR #29, opened 2023-10-19
fold in http_bindings + rpc into runtime

barraguda:
wip:
- paths to full processId paths [x]
- todo delete http_bindings + rpc? [x]

status: works without http_bindings or rpc being loaded in!


src/types.rs
```diff
@@ -727,6 +727,8 @@ pub enum HttpServerError {
         json
     )]
     BadJson { json: String, error: String },
+    #[error("http_server: path bindning error:  {:?}", error)]
```
willbach: spelling of `bindning`, should be `binding`

barraguda:
@willbach I made the change to bound paths so /processId/ is prepended to whatever path is bound. I changed some of the module html files to point at correct paths, but the javascript for example in chess needs to do the same.

summary: 
- chess binds "/" and "/games" instead of "/chess" and "/chess/games".
- these end up in bound paths as "/chess:chess:uqbar/" and "/chess:chess:uqbar/games" so that's the paths frontends will need to hit. they still come into processes with proper paths "/" and "/games"!

barraguda:
okay, this will be merged along with the change to chess-ui and package-template, there's one more thing I want to do, which is spawn a thread for every request (we'll probably need a queue per app).

dr-frmr:
Love the phat negative diff here.. code looks good, I haven't tested yet but excited to do so

dr-frmr:
Made a commit to remove the final compilation warnings and clean up the handle_response logic in http_server just a bit. Tested, merging!

PR #34, opened 2023-10-24
Jf/invite

dr-frmr:
this is NOT ready for merge. Network keys parsing is unreliable, not even deterministic it seems


mesozoic-technology:
Hmm very interesting. I hadn't tested this with any apps running after the registration process or login. 

I wonder how this is? When a keyfile is present but I adjusted the networking key on chain to be different, the login process was able to tell it didn't match what's on chain. If I didn't fudge the on chain value, it was okay.

At what point is this happening?

mesozoic-technology:
Now correctly formatting our.network_key to include prepended 0x

PR #38, opened 2023-10-25
vfs clean

barraguda:
currently: 
- prepends / appends '/' to files or directories that don't have them
- returns template errors instead of panics

todo:
- revamp types (specifically errors) to include inner strings, return them intelligently [x]
- implicit directory creation with /newdir1/newfile1.txt creates newdir1 if it doesn't exist, should probably work for nested things too [x]


src/vfs.rs
```diff
@@ -70,31 +70,96 @@ impl Vfs {
...
-        String::new()
-    } else {
-        format!("{}/", split_path.join("/"))
+#[async_recursion::async_recursion]
+async fn rename_entry(
+    vfs: Arc<Mutex<Vfs>>,
+    old_path: &str,
+    new_path: &str,
+) -> Result<(), VfsError> {
+    let (_key, children) = {
```
nick1udwig: Remove `_key` here and below at 148.

src/vfs.rs
```diff
@@ -70,31 +70,96 @@ impl Vfs {
...
+        let entry = match vfs.key_to_entry.get_mut(&key) {
+            Some(entry) => entry,
+            None => return Err(VfsError::EntryNotFound),
+        };
+        entry.name = new_path.split("/").last().unwrap().to_string();
+        entry.full_path = new_path.to_string();
+        let children = if let EntryType::Dir { children, .. } = &entry.entry_type {
+            children.clone()
+        } else {
+            HashSet::new()
```
nick1udwig, line 146: Change to `return Ok(())`

src/vfs.rs
```diff
@@ -70,31 +70,96 @@ impl Vfs {
...
-fn make_file_name(full_path: &str) -> (String, String) {
-    let mut split_path: Vec<&str> = full_path.split("/").collect();
-    let name = split_path.pop().unwrap_or("").to_string();
-    let path = if split_path.is_empty() {
-        String::new()
-    } else {
-        format!("{}/", split_path.join("/"))
+#[async_recursion::async_recursion]
+async fn rename_entry(
+    vfs: Arc<Mutex<Vfs>>,
```
nick1udwig, line 126: Consider changing `rename_entry()` to take a `MutexGuard` like in `create_entry()`: create code is more readable than rename.

src/vfs.rs
```diff
@@ -891,49 +864,27 @@ async fn match_request(
...
-                        panic!("");
+                        return Err(VfsError::InternalError);
                     };
                     let Some(mime) = payload.mime else {
-                        panic!("");
+                        return Err(VfsError::InternalError);
                     };
                     if "application/zip" != mime {
-                        panic!("");
+                        return Err(VfsError::InternalError);
```
nick1udwig: Are these appropriate uses of InternalError? Seems like this is a user input error, not an InternalError?

src/vfs.rs
```diff
@@ -1174,50 +1067,52 @@ async fn match_request(
...
+                        } = parent.entry_type
+                        else {
+                            return Err(VfsError::InternalError);
+                        };
+                        children.remove(&key);
+                    }
+                },
+            }
+            match persist_state(send_to_persist, &mut recv_response, id).await {
+                Err(_) => return Err(VfsError::PersistError),
```
nick1udwig: This and other calls to persist_state() seem like they can just `?` since persist_state() returns PersistError itself.

src/vfs.rs
```diff
@@ -1245,28 +1140,85 @@ async fn match_request(
...
+                return Err(VfsError::InternalError);
             };
+
             let Ok(FsResponse::Write(_)) =
                 serde_json::from_slice::<Result<FsResponse, FsError>>(&ipc).unwrap()
             else {
-                panic!("");
+                return Err(VfsError::InternalError);
+            };
+            match persist_state(send_to_persist, &mut recv_response, id).await {
```
nick1udwig: Do we actually need to persist_state here?

src/vfs.rs
```diff
@@ -1245,28 +1140,85 @@ async fn match_request(
...
+            };
+
+            let Ok(FsResponse::Append(_)) =
+                serde_json::from_slice::<Result<FsResponse, FsError>>(&ipc).unwrap()
+            else {
+                return Err(VfsError::InternalError);
             };
-            persist_state(send_to_persist, &mut recv_response, id).await;
-            (serde_json::to_vec(&VfsResponse::Ok).unwrap(), None)
+            match persist_state(send_to_persist, &mut recv_response, id).await {
```
nick1udwig: Same as in WriteOffset case: do we really need this?

src/vfs.rs
```diff
@@ -1295,16 +1246,21 @@ async fn match_request(
...
-                panic!("");
+                return Err(VfsError::InternalError);
             };
-            assert_eq!(size, length);
-            persist_state(send_to_persist, &mut recv_response, id).await;
-            (serde_json::to_vec(&VfsResponse::Ok).unwrap(), None)
+            if length != size {
+                return Err(VfsError::InternalError);
+            };
+            match persist_state(send_to_persist, &mut recv_response, id).await {
```
nick1udwig: Same as in WriteOffset case: do we really need to persist?

src/vfs.rs
```diff
@@ -70,31 +70,96 @@ impl Vfs {
...
-    (name, path)
+
+    // recursively update the paths of the children
+    for child_key in children {
+        let child_entry = {
+            let vfs = vfs.lock().await;
+            vfs.key_to_entry.get(&child_key).unwrap().clone()
+        };
+        let old_child_path = child_entry.full_path.clone();
+        let new_child_path = old_child_path.replace(old_path, new_path);
```
nick1udwig, line 160: Has this been tested? It looks to me like this will not update the full_path stored in the entry.

nick1udwig (changes requested):
Looks a lot nicer than before. Left a few comments.

For future work (not this PR), would be nice to update names internal to vfs to match better: some have been left behind. E.g., within we refer to the "hash" of a file in fs, but those are no longer hashes. Another example: we now refer to "vfss" as "drives".

src/vfs.rs
barraguda: there is probably a good way to do this that I've yet to figure out, but currently I need this because there's a loop of recursion within the recursion itself, every affected rename entry might have children itself

src/vfs.rs
barraguda: this is tested and works (now), but there's one thing left which I wonder how I should handle, the case where:

rename `/dir1/dir2/note` to `/dir1/dir2/dir3/note2`, or other implicit moving a file to another directory. I think we might want to limit renames to specific items and restrict moves, even though we could recursively loop and rename/create where needed

nick1udwig: The case you mention is not a rename imo: that's a move. A rename is `/foo/bar/` -> `/foo/baz/` or `/foo/bar.txt` -> `/foo/lol.txt`.

nick1udwig:
I'm going to do a quick test and then will approve.

src/vfs.rs
nick1udwig: That's fine, leave as is.

src/vfs.rs
nick1udwig: Confirmed this works.

src/vfs.rs
```diff
@@ -616,7 +686,6 @@ async fn handle_request(
     )
     .await?;
 
-    //  TODO: properly handle rsvp
     if expects_response.is_some() {
```
nick1udwig: This should be updated to work properly if rsvp is set but expects_response is not.

nick1udwig: Also: it would probably be good to print_to_terminal verbosity 1 the ipc in the case that no response is sent.

nick1udwig (approved):
Please address last two comments added just now, then ready for merge.

Really nice work, thanks @bitful-pannul !

src/vfs.rs
barraguda: cheers!

PR #40, opened 2023-10-26
modules: update wit and uq:process

barraguda:
update wit and microkernel_process to uq:process where applicable


dr-frmr:
ty, surprised to see this --

dr-frmr (approved):

PR #41, opened 2023-10-26
Uqbar 0.2.0

dr-frmr:
Change to wit file and thus the system at large: `ipc` is now a byte vector, not an Option<String>

@hosted-fornet I made changes in key_value and sqlite as unintrusive as possible, but you may want to rearrange types in light of this change. They are compiling now but I'm not sure if they're working. Would appreciate review to that end before we merge this

@bitful-pannul I will update the package template to reflect this change once we merge it here.

NOTE that this is a breaking change and will need all network nodes to upgrade to this version
ALSO make sure to delete apps / filesystem so new version of system apps get bootstrapped in


nick1udwig:
Pushed 1895804 and https://github.com/uqbar-dao/redb/commit/8e192d945295e1ddd331003bae215cd8d1564c4b and redb & sqlite now work for me :+1:

barraguda (approved):
added slight rpc change to server, pkg-template works with wit-change:
https://github.com/uqbar-dao/uqbar-package-template/pull/7

PR #42, opened 2023-10-31
start from commandline

nick1udwig:
This PR changes how we parse commandline flags -- see README.

Using clap to parse args lets us more readably parse args, as well as optional args, such as --password, which allows bypassing the FE sign in screen.

As of 1648962 seems to work as expected, but always sets us as an indirect node. Going to wait for more registration stability before trying to get this right. Please be aware of this usecase when you are working on registration: we want some sort of persisted routing flag that lets us know whether we are direct or indirect just based on Keyfile @operarotas @dr-frmr


dr-frmr:
In order to fully bypass the registration frontend, we will need the runtime to be able to call out to RPC and get the QNS identity info for the node it's trying to boot. If we don't have this, we will run into issues where identities change onchain but not on disk, so critical to always check -- this means this is somewhat blocked on having good RPC-gossip between nodes

dr-frmr:
This might be outdated with Will's recent registration flow work. I think we are using clap to parse args now too.

nick1udwig:
Yeah its going to be either a giant merge or a new PR, but this has evolved far beyond just command line arguments, and has all the test-related core changes.

nick1udwig:
For the record, the associated repos are
https://github.com/uqbar-dao/core_tests
https://github.com/uqbar-dao/uqdev

README.md
```diff
@@ -32,7 +32,7 @@ Get an eth-sepolia-rpc API key and pass that as an argument. You can get one for
 
 Make sure not to use the same home directory for two nodes at once! You can use any name for the home directory: here we just use `home`.
 ```bash
-cargo +nightly run --release home --rpc wss://eth-sepolia.g.alchemy.com/v2/<your-api-key>
+cargo +nightly run --release -- --home home --rpc wss://eth-sepolia.g.alchemy.com/v2/<your-api-key>
```
dr-frmr: what's going on here with the extra `--`?

src/kernel/mod.rs
```diff
@@ -924,6 +927,40 @@ pub async fn kernel(
                             content: format!("event loop: got message: {}", kernel_message)
                         }
                     ).await;
+
+                // send every single event to tester
+                // TODO: assess if performance will allow this
```
dr-frmr: has this been assessed? is this piece of code getting used ever or was it just a test thing?

README.md
nick1udwig: Its a separator; before it cargo is getting passed flags; after it, the binary is getting passed flags. If you compile the binary and run it you'd do
```
--home home --rpc wss://eth-sepolia.g.alchemy.com/v2/<your-api-key>
```
for equivalent flags.

src/kernel/mod.rs
nick1udwig: It is pretty slow and its not obvious to me how to speed it up. Also, I'm not certain this is a feature we want. It was in the spec that I described though.

nick1udwig: The idea is that you may, in principle, want to assert about things other than just the "start" and "end" of a chain of Req/Resp as your particular test process sees it.

src/types.rs
```diff
@@ -16,6 +17,82 @@ lazy_static::lazy_static! {
...
+
+    /// First port to try binding
+    #[arg(long, default_value_t = 8080)]
+    pub port: u16,
+
+    // /// Networking password
+    // #[arg(short, long, value_parser)]
+    // pub password: Option<String>,
+    /// LLM endpoint
+    #[arg(required(true), short, long, value_parser)]
```
dr-frmr: password flag shouldn't be required for LLM extension methinks

src/main.rs
```diff
@@ -185,39 +228,111 @@ async fn main() {
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        Some(name) => {
+            let password = match args.password {
+                None => "123".to_string(),
```
dr-frmr, line 335: also if the password flag is required, should .expect() it / not have a default value hardcoded

src/types.rs
nick1udwig: This was sloppy coding on my part; should've been `llm` not `password` here. As of 355befa, we now use clap dot instead of clap derive anyways and this is now properly set.

src/main.rs
nick1udwig: See comment above.

dr-frmr (approved):

PR #48, opened 2023-11-08
move process_lib to https://github.com/uqbar-dao/process_lib

nick1udwig:


dr-frmr (approved):
Assuming this builds and works as expected? I wasn't 100% certain that the runtime itself does not depend on the process_lib folder, looks like it!

PR #53, opened 2023-11-13
sqlite transactions

barraguda:
~~This is still in Draft phase.~~
we good.

Reason being, how rusqlite handles lifetimes in it's structs.

Essentially, when creating a transaction, it's expected it doesn't "outlive" the original Connection it references. Usually fine, if code looks something like this:

let conn = create_connection();
let tx = conn.tx()

tx.do_things()
tx.commit()
return, tx goes out of scope. 

In our case, we might want the tx to span several received messages, so it should be stored in mem.

**Things I'm trying**
refcell/arc
explicit lifetimes &'a
new connection every message?

### DIFF
--- a/modules/sqlite/sqlite_worker/src/lib.rs
+++ b/modules/sqlite/sqlite_worker/src/lib.rs
@@ -351,9 +351,10 @@ pub extern "C" fn send_and_await_response_wrapped(
     CIpcMetadata::copy_to_ptr(return_val, ipc, metadata);
 }
 
-fn handle_message<'a>(
+fn handle_message(
     our: &wit::Address,
-    db_handle: &'a mut Connection<'a>,
+    conn: &mut Option<rusqlite::Connection>,
+    txs: &mut HashMap<u64, rusqlite::Transaction>,
 ) -> anyhow::Result<()> {
     let (source, message) = wit::receive().unwrap();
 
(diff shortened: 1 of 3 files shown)


barraguda:
One cannot really sneak past the lifetime requirements, so I still use the Transactions struct, but only on Commit. Before that the statements and parameters accumulate in a memory hashmap. Worth thinking about limits on this mem-cache, which usually can be configured:

https://www.sqlite.org/lang_transaction.html
, configurable from within rusqlite too

barraguda:
also, overview of simple usage:
https://gist.github.com/bitful-pannul/06db8b953ac7ee009902b7892a36cd47

modules/sqlite/sqlite_worker/src/lib.rs
```diff
@@ -362,58 +364,75 @@ fn handle_message (
...
+                            *conn = Some(rusqlite::Connection::open_with_flags_and_vfs(
+                                    format!(
+                                        "{}:{}:/{}.sql",
+                                        our.node,
+                                        vfs_drive,
+                                        db,
+                                    ),
+                                    flags,
+                                    "uqbar",
+                                )?);
```
nick1udwig: Why the whitespace change? If not intentional should probably go back to what it was.

nick1udwig (approved):
Other than minor style nit in comment, LGTM. Nice work & good solution to the lifetime problems you were running into.

modules/sqlite/sqlite_worker/src/lib.rs
barraguda: oh.. auto-formatter :D

PR #60, opened 2023-11-14
fs: better dedup

barraguda:
I'm investigating a phenomenon where sqlite does way more writes than eventually end up on disk (10k inserts -> 300mb writes -> ends up as ~5mb deduped)

this leads me to think there's a lot of double writing. 

currently we dedupe on the cold-storage level, but there's no real reason to not do it on wal/mem level as well, as this pr does.

one thing I'm still investigating is whether I should do this for unconfirmed txs as well, but need to be careful.


barraguda:
merging, but doing a follow up PR that will address this at least on the memory buffer level. Treating memory buffer the same as WAL is useful for quick flushing and low complexity, but not good for writing the same chunks over and over.

PR #62, opened 2023-11-14
JSON interface for SQLite

willbach:
@bitful-pannul can you please test this out and make sure it works? It is currently compiling but untested.


barraguda:
refactored a bit and added json parameter injection to reads as well. works, I'll add test app example

barraguda:
updated example: 
https://gist.github.com/bitful-pannul/06db8b953ac7ee009902b7892a36cd47

barraguda (approved):
lgtm

PR #72, opened 2023-11-20
allowed unused mut

tadad:
`mut` is needed if `llm` feature is enabled. If not then you get a warning that you don't need the `mut` because the `llm` code doesn't get compiled. Added a line of code to bypass this warning so that no one gets rid of it


dr-frmr:
Ah, I see, sorry.

PR #75, opened 2023-11-24
LLM types changed

tadad:
Specifying & correcting types for llm

Removed a debug println


dr-frmr:
@tadad is this useful or should we close?

tadad:
let's close

PR #78, opened 2023-11-29
Wg/vector db

tadad:


dr-frmr:
@tadad @willbach any reason to keep this open?

tadad:
no this was just an uqville thing. Though we'll probably have to revive the vectordb runtime-ext with Nick's work at some point

PR #91, opened 2023-12-08
fs 2.0

barraguda:
- isolated kernel state in rocksdb + snapshots [x] 
- vfs -> thin layer over tokio::fs [x]
     - wasi/wnfs?
- kv & sqlite into runtime [x] [x]

also, uqdev/~~pkg-template~~/app_store to use new paths for installing packages! [x]


modules/app_store/app_store/src/lib.rs
```diff
@@ -317,12 +317,13 @@ fn handle_local_request(
             let Some(mut payload) = get_payload() else {
                 return Err(anyhow::anyhow!("no payload"));
             };
+            let drive = format!("/{}/{}", our.process.to_string(), package.to_string());
```
barraguda: this is something that's up for debate, currently drives are connected to process_ids, so `/process_id/drive_name/`is required before any path. 

This allows multiple drives per process, but also clutters up paths a little bit. Could do something more like we had before, 1 drive per package that's it, a bit restrictive.

modules/app_store/app_store/src/lib.rs
```diff
@@ -336,39 +337,36 @@ fn handle_local_request(
...
-                    },
+                    path: drive.clone(),
+                    action: kt::VfsAction::AddZip,
                 })?)
                 .payload(payload.clone())
                 .send_and_await_response(5)?.unwrap();
 
             // save the zip file itself in VFS for sharing with other nodes
             // call it <package>.zip
+            let zip_path = format!("{}/{}.zip", drive.clone(), package.to_string());
```
barraguda: ^example of formats done when writing a file with the full path

src/types.rs
```diff
@@ -151,6 +153,13 @@ impl ProcessId {
     pub fn publisher(&self) -> &str {
         &self.publisher_node
     }
+    pub fn to_hash(&self) -> [u8; 32] {
```
barraguda: used by state to key process_ids and their state, could be shorter than 32bytes?

src/types.rs
```diff
@@ -887,206 +894,140 @@ pub struct PackageManifestEntry {
...
+    ReadDir,
+    ReadExact(u64),
+    ReadToString,
+    Seek(SeekFrom),
+    RemoveFile,
+    RemoveDir,
+    RemoveDirAll,
+    Rename(String),
+    AddZip,
+    // Metadata,
```
barraguda: the file.metadata() operation seems to be quite common, returns a thing that looks something like this 
`
file metadata: Metadata { file_type: FileType(FileType { mode: 33188 }), is_dir: false, is_file: true, permissions: Permissions(FilePermissions { mode: 33188 }), modified: Ok(SystemTime { tv_sec: 1702565498, tv_nsec: 28397973 }), accessed: Ok(SystemTime { tv_sec: 1702565498, tv_nsec: 34742923 }), created: Ok(SystemTime { tv_sec: 1702565498, tv_nsec: 28293378 }), .. }
`

Could type it as a return string but not sure

src/vfs.rs
```diff
@@ -730,788 +444,245 @@ async fn handle_request(
...
+                .send(CapMessage::Has {
+                    on: source.process.clone(),
+                    cap: Capability {
+                        issuer: Address {
+                            node: our_node.clone(),
+                            process: VFS_PROCESS_ID.clone(),
+                        },
+                        params: serde_json::to_string(&serde_json::json!({
+                            "kind": "write",
+                            "drive": drive,
```
barraguda, line 570: drive in the capabilities is also represented by `/process_id/drive_name`

src/state.rs
```diff
@@ -0,0 +1,575 @@
...
+    let db = DB::open_default(state_path).unwrap();
+    let mut process_map: ProcessMap = HashMap::new();
+
+    let kernel_id = KERNEL_PROCESS_ID.to_hash();
+    match db.get(kernel_id) {
+        Ok(Some(value)) => {
+            process_map = bincode::deserialize::<ProcessMap>(&value).unwrap();
+        }
+        Ok(None) => {
+            bootstrap(
```
barraguda, line 42: this could/should be done by a simple process in main.rs/build that takes the .zips in the target folder (or in a specific distro folder), and extracts and puts them in the vfs folder within the node.

the reason why I kept it here, is that we need to add some state to the kernel about these modules + runtime extensions too, so either pass something in on-load or do it in the same place like this.

modules/app_store/app_store/src/lib.rs
barraguda: resolved to /package_id/drivename instead.

src/types.rs
barraguda: no need to hash, better with byte values for inspectability as well.

barraguda:
note: app_store changes use open pr version of process_lib:
https://github.com/uqbar-dao/process_lib/pull/11

PR #95, opened 2023-12-12
add py

nick1udwig:
In order to add support for Python, we need to update dependencies. This PR does that. It is associated with https://github.com/uqbar-dao/process_lib/pull/7 and https://github.com/uqbar-dao/uqdev/pull/3


nick1udwig:
See https://github.com/hosted-fornet/py-chat for an example app you can build.

dr-frmr:
Nice! This cause any build issues in the runtime? Looks good to me...

dr-frmr (approved):
good fix in that error print

nick1udwig:
> Nice! This cause any build issues in the runtime? Looks good to me...

Builds for me :+1:

PR #111, opened 2023-12-22
gracefully exit in case of many SIG*s

nick1udwig:
Previously, we manually caught `Ctrl+C` and `Ctrl+D` passed into our process. Upon receipt, we gracefully exit by restoring terminal behavior and title to normal. However, SIG*s like SIGINT would bypass this and lead to a rekt terminal.

This PR adds a new task, `intercept_sigs`, to catch the variety of SIG*s that lead to an exit and are handleable. When these are received, we print what we received, and then do the terminal teardown and graceful exit.

This PR is also important for `uqdev boot-fake-node` and `run-tests`, since `uqdev` must tear down the node(s) it is running. We can sometimes, but not always, pass a `Ctrl+C` equivalent into the node (how we used to do it). Being able to send a SIGINT is more general, but only works properly with this PR.

To test, try normal `Ctrl+C` exit (works here and on main) and then try `sudo kill <pid>` (works only here; on main it will rekt your terminal).


nick1udwig:
rustfmt doing real bad on this one

dr-frmr:
> rustfmt doing real bad on this one

ah yeah i hate what it's doing here

Cargo.toml
```diff
@@ -55,21 +55,21 @@ rand = "0.8.4"
 reqwest = "0.11.18"
 ring = "0.16.20"
 rmp-serde = "1.1.2"
+rocksdb = { version = "0.21.0", features = ["multi-threaded-cf"] }
 route-recognizer = "0.3.1"
+rusqlite = { version = "0.30.0", features = ["bundled"] }
```
dr-frmr, line 60: this isn't actually from this PR is it?

src/intercept_sigs.rs
```diff
@@ -0,0 +1,29 @@
+use tokio::signal::unix::{signal, SignalKind};
```
dr-frmr: i think this file might make more sense as a function inside the terminal task. and the `recv()` tasks could be `select!`ed in the main terminal select, rather than spinning up a new task just for this at boot. it could still be its own file i suppose, but mostly want to just move this all under `terminal` purview

Cargo.toml
nick1udwig: Correct: I am just alphabetizing those deps. Sorry for looping that into this single commit.

src/intercept_sigs.rs
nick1udwig: Good point. Consolidated in 25ec21c. Let us hope rustfmt favors us.

dr-frmr (approved):
This looks great, merge at will

PR #112, opened 2023-12-22
Da/caps api change

tadad:
WIP NOT DONE


tadad:
Question: is there redundancy between `messaging_capabilities` `networking_capabilities` and `public`? Might be able to get rid of one of those unless I'm misunderstanding...

PR #123, opened 2023-12-29
Da/caps 2

tadad:
WIP


modules/tester/test_runner/src/lib.rs
```diff
@@ -133,10 +133,11 @@ impl Guest for Component {
...
 
-        wit::create_capability(
-            &ProcessId::new(Some("vfs"), "sys", "uqbar"),
-            &"\"messaging\"".into(),
-        );
+        // TODO fix
+        // wit::create_capability(
+        //     &ProcessId::new(Some("vfs"), "sys", "uqbar"),
+        //     &"\"messaging\"".into(),
+        // );
```
tadad: @hosted-fornet I need to move this creation into the manifest, is there a reason why `test_runner` isn't in the manifest?

tadad:
more wit changes led me to restart on #133

PR #130, opened 2024-01-03
Add websocket client functionality to http/client

willbach:


src/http/client.rs
```diff
@@ -19,6 +33,8 @@ pub async fn http_client(
     let client = reqwest::Client::new();
     let our_name = Arc::new(our_name);
 
+    let ws_streams: Arc<Mutex<WebSocketMap>> = Arc::new(Mutex::new(HashMap::new()));
```
dr-frmr: Try using `dashmap` for this, it's much more performant and easier to read -- see example of usage in src/net/mod.rs

src/http/client.rs
```diff
@@ -277,3 +579,174 @@ async fn make_error_message(
...
+        )
+        .await;
+        return;
+    };
+
+    // Close the stream. The stream is closed even on error.
+    match ws_sink.close().await {
+        Ok(_) => {}
+        Err(_) => {}
+    }
```
dr-frmr: can just do `let _ = ws_sink.close().await;`

src/http/client.rs
```diff
@@ -33,22 +49,308 @@ pub async fn http_client(
...
+                Response {
+                    inherit: false,
+                    ipc: serde_json::to_vec::<WebSocketClientAction>(
+                        &WebSocketClientAction::Open {
+                            url: url.to_string(),
+                            headers,
+                            channel_id,
+                        },
+                    )
+                    .unwrap(),
```
dr-frmr: Here, are we just sending the exact request that the process sent us back as a response? I don't think that's ideal -- what if we had a dedicated Response enum for the WebSocketClient that had all the errors and Ok results?

dr-frmr (changes requested):

dr-frmr:
In addition to review, general architecture note:

Spawning a task for each message is going to be costly vs. handling messages in a hot loop, and only spawning tasks when new long-lived connections need to be opened. But performance cost isn't the biggest issue. I haven't tested this code, but it looks like it will have a predictable issue where if two messages are received in rapid succession, two tasks will be spawned, and two websocket connections will be opened. The correct behavior would be to have the second message block until the first one resolves by opening a new connection. So, I think a better layout might be to have one main loop responsible for handling messages, and block until a message is either kicked out to an existing connection, or a new one is opened (or a failure occurs).

dr-frmr (approved):
Looks good!

PR #134, opened 2024-01-05
kernel restart change

barraguda:
So that we don't have to send messages with the `StartProcessMetadata` struct serialized to a string (problem with a caps HashMap<Struct, Vec<u8>>)


barraguda:
local and remote installing from app_store and restarting working with da/caps branches on wit and process_lib

PR #151, opened 2024-01-11
Add better debugging commands to term

dr-frmr:


dr-frmr:
@tadad I fiddle with the display for various types here -- hopefully simplified some of the newline logic, but it's still gross...

PR #158, opened 2024-01-15
http raw_path fix

barraguda:


src/http/server.rs
```diff
@@ -465,7 +465,14 @@ async fn http_handler(
             }
         }
     } else {
-        // otherwise, make a message to the correct app
+        // remove process id from beginning of path by splitting into segments
+        // separated by "/" and taking all but the first
+        let segments: Vec<&str> = original_path.split('/').collect();
+        let stripped_path = if segments.len() > 2 {
+            segments[2..].join("/")
```
willbach: you want to prepend a slash on this, I don't think it is rn

dr-frmr (approved):

PR #164, opened 2024-01-17
manifest: require *_capabilities

nick1udwig:
We should not let `request_capabilities` and `grant_capabilities` fields be optional in `manifest.json`. This PR makes them required fields.

There is no benefit to these being optional fields.

The cost is in terms of debugging. When they are required fields, applications that are incorrectly configured will "fail fast" and error in unconfusing ways.

What is an example of an error? We have already had multiple errors due to the name change from `request_messaging` to `request_capabilities`, so this is not a hypothetical. A counterpoint might be that "that is a one-time thing". Maybe it looks that way now, but this requirement makes the code fail faster so if, in future things change, things will fail in a reasonable way for users with out-of-date `manifest.json`s.

Another case is misspellings. What if a user has `request_capability`? Without this change, it could take a long time to track down the spelling error: it will look like capabilities are simply not being assigned for some unknown reason. With this change, the types will cause the misspelling to be clearly the error, saving dev time.

Linked with https://github.com/uqbar-dao/process_lib/pull/32


dr-frmr:
I like this change, there's no good reason to have these be optional.

dr-frmr (approved):

Issue #166, opened 2024-01-17
boot: using the first available port


dolled-possum:
Regarding the routine around line 200 in main.rs for failing if the port isn't open and recommending the next available port, this makes a lot of sense for when the --port arg is specified: the user intended one thing, but it couldn't be done, so an error messages suggests an alternative they could do that would work.  But as --port is optional, I think different behavior is appropriate if it's not specified, as it's not like the user was wedded to the choice of 8080: it was merely the behind the scenes default.  If they're not specifying a port, the user would likely be fine with whatever port was assigned (as long as they're informed what it is).  Can this functionality be amended to: "If --port isn't specified, use 8080 or the first available port higher (and print the port used in the terminal); otherwise (i.e., if --port is designated), use the existing logic."?

(I'm bringing this up for improved compatibility for automated multi-tenant hosting deployments.  While the hosting environment _could_ do this same logic, it's redundant as the nectar instance is already doing so in order to provide the suggestion.  It might as well just take its own suggestion!)

dr-frmr:
Thank you for the suggestion -- we *did* have this behavior originally, with automatic port selection, but we made the choice to lock in a port selection in the belief that it would make automated deployments and testing configurations easier.

If this isn't the case, then I see no reason why we shouldn't do what you're suggesting and only "enforce" a port if a selection was actually made. Seems like the best of both worlds to me...

dr-frmr:
I think the best mix is to have the port be freely assigned to an open one if the flag is not present, but if the flag is present, the runtime will refuse to boot unless the selected port is available. I'll write a PR for this

dr-frmr:
#168 should resolve this!

PR #168, opened 2024-01-17
Dr/port flag enforce only

dr-frmr:
Addresses issue #166


src/main.rs
```diff
@@ -211,16 +210,36 @@ async fn main() {
...
+                    "error: couldn't bind any ports between {port} and {}. \
+                    Set an available port with `--port` and try again.",
+                    port + 1000,
+                );
+                panic!();
+            }
+        }
+    };
+
+    if http_server_port != port {}
```
nick1udwig: What is this line doing?

dr-frmr: lol whoops!

PR #169, opened 2024-01-17
app_store: add sqlite & kv perms so they can be requested

nick1udwig:
Without this change, packages started by app_store cannot themselves request these caps.

Any others we should add here?


dr-frmr:
Ah yes, we will need to give app_store all the distro caps. Let's also give it messaging access to `chess:chess:sys` and `kns_indexer:kns_indexer:sys`

PR #173, opened 2024-01-17
Da/remove cap sigs

tadad:
- removes all unnecessary signature verification from capabilities
- changes `CapMessage::GetSome` to `CapMessage::FilterCaps`
- re-signs all capabilities on boot to fix any issues with invalid caps when booting after a key reset
  - note that since we don't check local signatures, this doesn't matter in 99% of cases. But without doing this, we would risk sharing an outdated cap-signature with a foreign node. Since it doesn't change boot-time much, I just re-sign every local cap on boot

### DIFF
--- a/src/kernel/mod.rs
+++ b/src/kernel/mod.rs
@@ -199,18 +199,8 @@ async fn handle_kernel_request(
             } else {
                 for cap in initial_capabilities {
                     match parent_caps.get(&cap) {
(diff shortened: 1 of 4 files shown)


dr-frmr:
@tadad is this all tested and ready to go?
does it break the disk-state of the kernel, i.e. can you upgrade to it from old version seamlessly?

tadad:
All tested and ready to go - no kernel state changes required

Issue #183, opened 2024-01-20
non "nerdview" errors


nick1udwig:
Errors should be targeted at the user of the software, not the developer of it. They should provide some sort of information about "what to do next" or "how to fix this problem". They should not be ["nerdview"](https://languagelog.ldc.upenn.edu/nll/?p=276).

E.g. in chat someone hit this error:
```
uh oh, a kernel process crashed -- this should never happen: Err(eth: fatal: http provider not supported yet!)
```

This is a nerdview error: `http provider not supported yet!` is useless to a user.

A better error would be:

```
You provided an `http://` Ethereum RPC, but only `wss://` is supported, please try again with a `wss://` provider.
```

If should probably also not output that scary-looking

```
uh oh, a kernel process crashed -- this should never happen
```

in a case like this that is just bad user input.

PR #184, opened 2024-01-20
eth: change two nerdview error messages

nick1udwig:
I've addressed the specific example mentioned in #183, but the general principle still applies.


dr-frmr (approved):
tyvm, merging

PR #185, opened 2024-01-20
app_store: fix `get_capability()` bug by using fixed process_lib

nick1udwig:
Fix the app store by using https://github.com/uqbar-dao/process_lib/pull/33

I don't have the time to go through the codebase and check right now, but someone needs to be assigned to:

1. Check through all capabilities comparisons to make sure they are not doing same improper equality check as in https://github.com/uqbar-dao/process_lib/pull/33
2. Make sure https://github.com/uqbar-dao/process_lib/pull/33 is a fix we're comfortable with (there may [should?] be a better way to do it)
3. Check through all processes to make sure that if they are using `get_capability()` (and any other affected functions found in 3) that they are pointing to a new enough `process_lib`.


dr-frmr (commented):
Changing to v0.5.5-alpha since I just released that

PR #186, opened 2024-01-22
Da/script runner

tadad:
this is the MVP version of scripts in kinode. #188 details the future improvements that can/will be made.


modules/terminal/terminal/src/lib.rs
```diff
@@ -11,87 +16,40 @@ wit_bindgen::generate!({
...
+    pub root: bool,
+    pub public: bool,
+    pub request_networking: bool,
+    pub request_capabilities: Option<Vec<serde_json::Value>>,
+    pub grant_capabilities: Option<Vec<serde_json::Value>>,
+}
+
 struct TerminalState {
     our: Address,
     current_target: Option<Address>,
```
tadad: NOTE: the `/a` functionality was removed - I think the way this should work in the future is being able to alias `FOO=foo.kino@bar:baz:jax` and then call `m FOO { foo: bar }`

modules/terminal/terminal/src/lib.rs
```diff
@@ -132,3 +114,192 @@ impl Guest for Component {
...
+    let wasm_path = format!("{}{}", drive_path, wasm_path);
+    // build initial caps
+    let mut initial_capabilities: HashSet<kt::Capability> = HashSet::new();
+    if entry.request_networking {
+        initial_capabilities.insert(kt::de_wit_capability(Capability {
+            issuer: Address::new(&our.node, ("kernel", "distro", "sys")),
+            params: "\"network\"".to_string(),
+        }));
+    }
+    let process_id = format!("{}:{}", rand::random::<u64>(), package); // all scripts are given random process IDs
```
tadad, line 141: note that we call scripts as if they were long running processes e.g. `hi:terminal:sys` but you want to be able to run multiple versions of the same script at once, so we use a random process ID under the hood

modules/terminal/terminal/src/lib.rs
tadad: deleted current_target formally

README.md
```diff
@@ -84,28 +84,32 @@ The `sys` publisher is not a real node ID, but it's also not a special case valu
...
 - CTRL+R to search history, CTRL+R again to toggle through search results, CTRL+G to cancel search
 
-- `/message <address> <json>`: send an inter-process message. <address> is formatted as <node>@<process_id>. <process_id> is formatted as <process_name>:<package_name>:<publisher_node>.
-    - Example: `/message our@net:distro:sys diagnostics`
+- `m <address> <json>`: send an inter-process message. <address> is formatted as <node>@<process_id>. <process_id> is formatted as <process_name>:<package_name>:<publisher_node>.
+    - Example: `m our@net:distro:sys diagnostics`
     - `our` will always be interpolated by the system as your node's name
-    - Can also use `/m` for same command: `/m our@net:distro:sys diagnostics`
-- `/app <address>`: set the terminal to a mode where all messages go to a specific app. To clear this selection, use `/app clear` or simply `/app`. This is useful for apps that have a command line interface.
+    - Can also use `m` for same command: `m our@net:distro:sys diagnostics`
```
dr-frmr: this can probably be removed due to changes above it, right?

README.md
```diff
@@ -84,28 +84,32 @@ The `sys` publisher is not a real node ID, but it's also not a special case valu
...
+<!-- - `/app <address>`: set the terminal to a mode where all messages go to a specific app. To clear this selection, use `/app clear` or simply `/app`. This is useful for apps that have a command line interface.
```
dr-frmr: what's going on here?

modules/terminal/echo/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+
+call_init!(init);
+
+fn init(_our: Address) {
+    let Ok(args) = await_next_request_body() else {
+        println!("echo: failed to get args, aborting");
+        return;
+    };
+
+    println!("{}", String::from_utf8(args).unwrap());
```
dr-frmr: maybe do unwrap_or here just so it never ever crashes? idk what happens when a script crashes

modules/terminal/hi/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        println!("hi: failed to get args, aborting");
+        return;
+    };
+
+    let tail = String::from_utf8(args).unwrap();
+
+    let (node_id, message) = match tail.split_once(" ") {
+        Some((s, t)) => (s, t),
+        None => {
+            println!("invalid command: \"{tail}\"");
```
dr-frmr: would be nice to give more clarity here about "supply more arguments, need <node_id> <message>"

modules/terminal/m/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+            println!("invalid address: \"{target}\"");
+            return;
+        } // match state.aliases.get(target) {
+          //     Some(pid) => Address::new("our", pid.clone()),
+          //     None => {
+          //         return Err(anyhow!("invalid address: \"{target}\""));
+          //     }
+          // },
+    };
+    let _ = Request::new().target(target).body(body).send();
```
dr-frmr: you actually *can* unwrap here, since request.send() will never fail if target and body exist

modules/terminal/terminal/src/lib.rs
```diff
@@ -100,7 +45,30 @@ impl Guest for Component {
...
+                a.insert(
+                    "m".to_string(),
+                    "m:terminal:sys".parse::<ProcessId>().unwrap(),
+                );
+                a.insert(
+                    "top".to_string(),
+                    "top:terminal:sys".parse::<ProcessId>().unwrap(),
+                );
+                a
+            },
```
dr-frmr: rather than doing a bunch of inserts, use `HashMap::from()` syntax here!

modules/terminal/top/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+
+call_init!(init);
+
+fn init(_our: Address) {
+    let Ok(args) = await_next_request_body() else {
+        println!("top: failed to get args, aborting");
+        return;
+    };
+
+    let proc_id = String::from_utf8(args).unwrap();
```
dr-frmr: let's return an understandable error here

modules/terminal/top/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    let kernel_addr = Address::new("our", ("kernel", "distro", "sys"));
```
dr-frmr: can declare this inline in request:
`.target(("our", (Some("kernel"), "distro", "sys"))`

modules/terminal/top/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+    let _ = Request::new()
+        .target(kernel_addr)
+        .body(
+            serde_json::to_vec(&match proc_id.parse::<ProcessId>() {
+                Ok(proc_id) => KernelCommand::Debug(KernelPrint::Process(proc_id)),
+                Err(_) => KernelCommand::Debug(KernelPrint::ProcessMap),
```
dr-frmr: i would rather we return an error when the given process ID doesn't parse, and only do a `KernelPrint::ProcessMap` in the case where they call `top` with no arg at all

dr-frmr (changes requested):

README.md
tadad: oh so I removed this because it causes some complication with the interaction between `m` and `terminal`. I could add it back, but I'm not sure it's the right solution. I think the right thing to do later would be to have aliases like `export FOO=our@net:distro:sys` and then calling `m FOO diagnostics`

modules/terminal/echo/src/lib.rs
tadad: when a script crashes you get a printout like this:

`Mon 1/22 12:12 process 17912472128299023401:terminal:sys ended with error:     wasm trap: wasm `unreachable` instruction executed`

dr-frmr (approved):

PR #193, opened 2024-01-22
change ssh import to https

nick1udwig:
Inspired by https://github.com/uqbar-dao/kit/issues/46

There is no reason we need to use `ssh://` now that we're public and it won't work for users that don't have their git set up in a certain way. Therefore, use `https://`.


dr-frmr (approved):
thanks for going through and doing this. i had been doing it lazily as i saw them, this is good to go

PR #196, opened 2024-01-23
Da/alias

tadad:


modules/terminal/terminal/src/lib.rs
```diff
@@ -81,14 +92,39 @@ impl Guest for Component {
...
+
+                        match edit_aliases.process {
+                            Some(process) => {
+                                state
+                                    .aliases
+                                    .insert(edit_aliases.alias.clone(), process.clone());
+                                println!(
+                                    "terminal: alias {:?} set to {:?}",
+                                    edit_aliases.alias,
+                                    process.to_string()
```
dr-frmr: instead of doing `process.to_string()` you can just print it without the `:?` debug, it will format automatically. same for alias i believe, no need to debug-print

modules/terminal/alias/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+                let _ = Request::new()
+                    .target(("our", "terminal", "terminal", "sys"))
+                    .body(
+                        json!({
+                            "alias": alias,
+                            "process": process
+                        })
+                        .to_string()
+                        .as_bytes()
+                        .to_vec(),
```
dr-frmr: I'd prefer to copy in the "real type" here and serialize it, makes it more upgradable later

dr-frmr (changes requested):
Two small nits then good to go

PR #201, opened 2024-01-24
add hi-protocol to mock networking

dr-frmr:


nick1udwig (approved):
It works, thanks for the fix :+1:

PR #202, opened 2024-01-24
add python code runner

nick1udwig:
# Goal

Enable an AI-assisted workflow for developing vanilla python code that looks like
1. Prompt LLM, e.g., "write a program that computes the fibonacci sequence",
2. LLM generates python code,
3. Run python code,
4. Take result and iterate python code until it is correct.

This PR enables 3. In particular, it is important that the python be sandboxed from the rest of the system, since it may be faulty and may even be malicious (in the case of prompt injection to the LLM). Note that this code correctly handles & returns error output from the python interpreter, which is vitally important for the proposed workflow.

# API

API is very simple:

https://github.com/uqbar-dao/kinode/blob/ae9fd4254860373cb4e00cacf2b080b3203295ab/src/types.rs#L1343-L1354

1. Send a `"Run"` `Request` with utf-8 bytes in the blob: those bytes are the python code.
6. Receive a `"Run"` `Response` with utf-8 bytes in the blob: those bytes are the stdout/stderr of the python code being run.

# Usage

Usage looks like:

https://twitter.com/nick1udwig/status/1750037861750210942

# Notes

We `include_bytes!()` the `python.wasm` in the runtime. This adds roughly 20MB to the size of the runtime. There are two ways we could reduce this:
1. Compress the `.wasm` before including it. This roughly halves the size.
2. Fetch the `.wasm` at runtime and store it in memory. This would probably look to the user like an initial "startup" period of the `python` runtime module.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ target.wasm
 world
 .env
 src/bootstrapped_processes.rs
+src/python_includes.rs
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2758,6 +2758,7 @@ dependencies = [
  "uuid 1.4.1",
  "walkdir",
  "warp",
+ "wasi-common",
  "wasmtime",
  "wasmtime-wasi",
  "zip",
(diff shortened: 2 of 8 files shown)


nick1udwig:
Closing this PR in favor of #218

PR #208, opened 2024-01-25
Da/ma

tadad:
`ma` stands for `message await` which is just `m` but adds an additional `u64` param so that you can await the response and print it out in the terminal

`ma 5 our@foo:bar:baz {"foo":"bar}`

NOTE
this is a somewhat unsatisfactory solution, originally I wanted to use `clap` to parse a new optional `--await` argument, but unfortunately clap doesn't play well with json args. Internally, it splits everything up by spaces, so a json body arg that uses spaces would screw everything up (e.g. `{"foo": "bar"}` would split into `{"foo":` and `"bar"}`. I tried a few other configurations but making a new `ma` script ended up being the easiest.


nick1udwig:
What about passing in the json as single quoted, e.g. `'{"hello": "yes", "json": ["lol"]}'`? This is how we do it in `kit inject-message`.

tadad:
i need to find the code that they are using for their CLI parsing  - for us, the args aren't coming through stdin - they're coming via a message that I then split based on whitespace and then pass to clap

dr-frmr (approved):

tadad:
@hosted-fornet I got your suggestion to work in #210, going to close this one!

PR #215, opened 2024-01-29
Da/pipe

tadad:
let's you compose scripts via piping.
If your script can be composed, you will finish it by sending a `Response` to the terminal. If the pipe is set, it will feed that output into the next script, otherwise, it will print out the `res.body()`. If the script does not send a `Response` at the end, then any attempt to pipe it's output will timeout.

Piping is done via `|<DIGIT>` e.g. `|5` or `|1000` where the digit represents the time to await the first script before timing out.

Examples:
```bash
echo default-router-1.os Hello! |1 hi
``` 
takes the output of the first command, `echo default-router-1.os Hello!` and pipes it into `hi`, with a timeout set for one second on echo to respond before aborting


dr-frmr (approved):
Ok fine, ONE last thing before 0.5.2!

PR #222, opened 2024-02-01
new eth

barraguda:
goals: 
- remove ethers-rs in favor of alloy-rs [x]
- signatures, easy calls from processes (send-and-await) [x]
- rpc passthrough [x]
- remove ethers from src/register/client [x]


kinode/src/main.rs
```diff
@@ -24,7 +24,7 @@ mod state;
 mod terminal;
 mod timer;
 mod vfs;
-
+//
```
dr-frmr: delet

kinode/src/kernel/mod.rs
```diff
@@ -826,6 +845,82 @@ pub async fn kernel(
         })
         .await
         .expect("fatal: kernel event loop died");
+    // sending hard coded routers into networking for bootstrapped rpc
+    let kns_updates = vec![
```
dr-frmr: We need to move this elsewhere. Possibly into a dedicated JSON file that gets parsed in main? Not here though

kinode/src/kernel/mod.rs
```diff
@@ -826,6 +845,82 @@ pub async fn kernel(
...
+            node: "0x".to_string(),
+            public_key: "0xf8a3e9667756306a0a25894a8cfe053dc9e7f34e6a61b8d65e92b79f77e05dff"
+                .to_string(),
+            ip: "this should not be parsed.".to_string(),
+            port: 0,
+            routers: vec![
+                "0xb35eb347deb896bc3fb6132a07fca1601f83462385ed11e835c24c33ba4ef73d".to_string(),
+                "0xd827ae579fafa604af79fbed977e8abe048497f10885c6473dfd343a3b7b4458".to_string(),
+                "0x96e36331c8f0882f2c0c46c13b15d812def04fe8606d503bc0e2be39db26486a".to_string(),
+            ],
```
dr-frmr: Remove when possible, of course

dr-frmr (commented):
This looks really good so far. I need to run locally and play around with it, but looks almost ready to merge.

kinode/src/kernel/mod.rs
barraguda: moved it into kinode/default-routers.json

it attempts to read home/.routers first though. and this is only passed to net:distro:sys, so if the user has any newer state for these routers in the indexer, they should be overwritten correctly on boot

kinode/packages/app_store/app_store/src/lib.rs
```diff
@@ -174,11 +185,11 @@ fn handle_message(
             Req::FTWorkerResult(r) => {
                 println!("app store: got weird ft_worker result: {r:?}");
             }
-            Req::Eth(e) => {
+            Req::Eth(sub) => {
                 if source.node() != our.node() || source.process != "eth:distro:sys" {
                     return Err(anyhow::anyhow!("eth sub event from weird addr: {source}"));
                 }
```
barraguda: note here: with current passthrough, this would hit wall at source != our

kinode/src/main.rs
```diff
@@ -377,6 +387,24 @@ async fn main() {
...
+                let routers: Vec<KnsUpdate> = serde_json::from_str(DEFAULT_ROUTERS).unwrap();
+                // should we dump them into a .routers file?
+                // fs::write(
+                //     format!("{}/.routers", home_directory_path),
+                //     serde_json::to_string(&routers).unwrap(),
+                // )
+                // .await
+                // .unwrap();
+                routers
+            }
```
barraguda: also left this commented out as a question, probably going to remove it, letting user specifically write .routers if necessary. perhaps should have a specific eth-rpc node config too?

kinode/packages/app_store/app_store/src/lib.rs
dr-frmr: If we remove this assertion, we could receive fraudulent updates from just about anyone -- if the passthrough is in the runtime, we should manipulate the `source` field to look like it's us when we get a good passthrough message.

kinode/src/main.rs
dr-frmr: I don't think we should do this. After first boot, we don't need to bootstrap from hardcoded PKI info, since we persist it when we get it.

lib/src/eth.rs
```diff
@@ -1,52 +1,97 @@
...
+    },
     /// Kill a SubscribeLogs subscription of a given ID, to stop getting updates.
     UnsubscribeLogs(u64),
+    /// Raw request. Used by kinode_process_lib.
+    Request {
+        method: String,
+        params: serde_json::Value,
+    },
+}
+/// Incoming subscription update.
```
dr-frmr: add a whitespace here, and say that this is the incoming Request type that processes will receive from eth. Same change in process_lib PR

lib/src/eth.rs
```diff
@@ -1,52 +1,97 @@
...
+        // "net_peerCount" => Some("net_peerCount"),
+        // "net_listening" => Some("net_listening"),
+        // "web3_clientVersion" => Some("web3_clientVersion"),
+        // "web3_sha3" => Some("web3_sha3"),
+        _ => None,
+    }
+}
+
+pub enum ProviderInput {
+    WS(String),
```
dr-frmr: uncapitalize the S in `WS`, to stick with convention in rest of repo

kinode/src/main.rs
```diff
@@ -45,11 +46,13 @@ const VERSION: &str = env!("CARGO_PKG_VERSION");
 #[cfg(not(feature = "simulation-mode"))]
 const REVEAL_IP: bool = true;
 
+/// default routers as a fallbac
```
dr-frmr: fallback

kinode/src/kernel/mod.rs
```diff
@@ -671,6 +672,7 @@ pub async fn kernel(
     home_directory_path: String,
     contract_address: String,
     runtime_extensions: Vec<(t::ProcessId, t::MessageSender, bool)>,
+    routers: Vec<crate::net::KnsUpdate>,
```
dr-frmr: need to rename this variable to `bootstrapped_pki_entries` or something clearer. in general, this new code needs to distinguish between "routers" as a concept and what's going on here, which is unrelated to network-routing

kinode/packages/kns_indexer/pkg/manifest.json
```diff
@@ -3,7 +3,7 @@
         "process_name": "kns_indexer",
         "process_wasm_path": "/kns_indexer.wasm",
         "on_exit": "Restart",
-        "request_networking": false,
+        "request_networking": true,
```
dr-frmr: no longer need this change due to source-spoof correct?

kinode/src/eth/provider.rs
```diff
@@ -1,189 +1,471 @@
...
-                    expects_response: None,
-                    body: serde_json::to_vec(&EthSubEvent::Log(event)).unwrap(),
-                    metadata: None,
-                    capabilities: vec![],
-                }),
-                lazy_load_blob: None,
-            })
-            .await
-            .unwrap();
+// todo, always send errors or no? general runtime question for other modules too.
```
dr-frmr: only send error if response is expected

dr-frmr (changes requested):

kinode/src/eth/provider.rs
barraguda: ok, I think I should make the case of a closed subscription special, as a process probably would like to know that even if subs have no expects_response

dr-frmr: well yes, that's a Request isn't it

dr-frmr (approved):
LFG

Issue #220, opened 2024-01-30
scripts: improve error handling


nick1udwig:
One specific comment and one general comment.

## Specific comment

`alias` can be used in unexpected ways that would ideally error out differently, e.g.:

### Alias can be assigned to any process; only fails when that alias is called

Current behavior:
```
fake.os > alias echo echo:echo:sys
Tue 1/30 10:01 terminal: alias echo set to echo:echo:sys
fake.os > echo hi
Tue 1/30 10:01 terminal: failed to instantiate script: couldn't find /echo:sys/pkg/scripts.json
```
```
fake.os > alias echo app_store:app_store:sys
Tue 1/30 10:01 terminal: alias echo set to app_store:app_store:sys
fake.os > echo
Tue 1/30 10:01 terminal: failed to instantiate script: script not in scripts.json file
```

Desired behavior:
Error when the alias is set and provide non-"nerdview" errors. E.g. in first case it'd be nice to say something like "package `echo:sys` does not exist" and in the second something like "given process is not a script; can only alias scripts"

### Different error print if alias DNE

Current behavior:
```
fake.os > alias echo
Tue 1/30 09:59 terminal: alias echo removed
fake.os > alias echo
Tue 1/30 09:59 terminal: alias echo removed
```

Desired behavior:
Print `alias does not exist` if an alias does not exist.

## General comment

It'd be nice if scripts printed out usage information if they can't parse args. Some are currently more helpful than others, but in general it'd be great to print usage if given no args. E.g., we use `clap` for some of the arg parsing: can we get that usage information for those that we do?

Current behavior:
```
fake.os > m
Tue 1/30 10:03 m: failed to parse args
fake.os > alias
Tue 1/30 10:04 alias: no alias given
```

FYI @tadad

nick1udwig:
> we use clap for some of the arg parsing: can we get that usage information for those that we do?

See, e.g.,
https://github.com/kinode-dao/kit/blob/ed92ceb8e3026c865fbf2746dfcbafaaeed4ae9b/src/main.rs#L609
https://docs.rs/clap/latest/clap/struct.Command.html#method.render_usage

nick1udwig:
#234 is a good step towards error handling. However, we should print usage whenever we get bad input, not just for empty input.

[...]

PR #234, opened 2024-02-07
Da/drop

tadad:
NOTE: can't merge before https://github.com/kinode-dao/process_lib/pull/50 (need to update `Cargo.toml`s to reflect latest version)

- ~~`drop_capabilities` added to `standard_host.rs`~~
- Bug fixed with `OnExit::Requests`
- All granted caps are cleaned up when a process is killed
- addresses #220 with better error handling in `alias`

Note that while `DropCaps` isn't used in the final version of this PR, it was tested on an earlier commit


Cargo.lock
```diff
@@ -3191,8 +3191,8 @@ dependencies = [
 
 [[package]]
 name = "kinode_process_lib"
-version = "0.5.9"
-source = "git+https://github.com/uqbar-dao/process_lib.git?tag=v0.5.9-alpha#5e705086bbd10fde89e11d3e3671f6a618a875a7"
+version = "0.5.8"
+source = "git+https://github.com/uqbar-dao/process_lib.git?rev=9668c82#9668c820986e32f7b8430648147c2498d5abb9f8"
```
dr-frmr: Cargo.lock has reverted somewhat, make sure to target the latest version of process_lib once you merge the matching PR there.

kinode/packages/terminal/alias/src/lib.rs
```diff
@@ -38,7 +40,7 @@ fn init(_our: Address) {
         let _ = Request::new()
             .target(("our", "terminal", "terminal", "sys"))
             .body(
-                json!(EditAliases {
+                json!(TerminalAction::EditAlias {
```
dr-frmr: Always better to use `serde_json::to_vec`, instead of using the macro and then double-copying with to_string() then to_vec()

kinode/packages/terminal/alias/src/lib.rs
```diff
@@ -53,7 +55,7 @@ fn init(_our: Address) {
                 let _ = Request::new()
                     .target(("our", "terminal", "terminal", "sys"))
                     .body(
-                        json!(EditAliases {
+                        json!(TerminalAction::EditAlias {
```
dr-frmr: serde_json::to_vec

kinode/packages/terminal/terminal/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 [dependencies]
 anyhow = "1.0"
 bincode = "1.3.3"
-kinode_process_lib = { git = "https://github.com/kinode-dao/process_lib", tag = "v0.5.9-alpha" }
+kinode_process_lib = { git = "https://github.com/uqbar-dao/process_lib.git", rev = "9668c82" }
```
dr-frmr: Use `kinode-dao` for URL and use tagged version of process_lib after merging PR there

kinode/packages/terminal/terminal/src/lib.rs
```diff
@@ -205,7 +204,7 @@ fn handle_run(
     };
     let wasm_path = format!("{}{}", drive_path, wasm_path);
     // build initial caps
-    let process_id = format!("{}:{}", rand::random::<u64>(), package); // all scripts are given random process IDs
+    let process_id = format!("{}:terminal:sys", rand::random::<u64>()); // all scripts are given random process IDs
```
dr-frmr: is this correct? shouldn't a script from another package use its package name?

kinode/packages/terminal/terminal/src/lib.rs
```diff
@@ -205,7 +204,7 @@ fn handle_run(
...
     let Ok(parsed_new_process_id) = process_id.parse::<ProcessId>() else {
         return Err(anyhow::anyhow!("app store: invalid process id!"));
```
dr-frmr: this code is in terminal, error print should reflect that

kinode/packages/terminal/terminal/src/lib.rs
```diff
@@ -396,3 +410,67 @@ fn handle_run(
...
+                return Err(anyhow::anyhow!(
+                    "couldn't find /{}/pkg/scripts.json",
+                    process.package()
+                ));
+            };
+            let dot_scripts = String::from_utf8(blob.bytes)?;
+            let dot_scripts =
+                serde_json::from_str::<HashMap<String, kt::DotScriptsEntry>>(&dot_scripts)?;
+            let Some(_) = dot_scripts.get(&format!("{}.wasm", process.process())) else {
+                return Err(anyhow::anyhow!("script not in scripts.json file"));
```
dr-frmr: i actually don't know if this check should be done. if someone sets an alias to a nonexistent script, is the failure at the time of using the alias not clear enough? it should indicate the underlying script does not exist -- and it's possibly nice to be able to set aliases before actually installing a script. open to thoughts here

kinode/packages/terminal/terminal/src/lib.rs
```diff
@@ -396,3 +410,67 @@ fn handle_run(
...
+                println!("terminal: alias {} not found", alias);
+            }
+        }
+    }
+    if let Ok(new_state) = bincode::serialize(&state) {
+        set_state(&new_state);
+        Ok(())
+    } else {
+        Err(anyhow!("failed to serialize state!"))
+    }
```
dr-frmr: this can be productively compressed into a one-liner with `?`, also a very very very rare error

kinode/packages/terminal/terminal/src/lib.rs
```diff
@@ -396,3 +410,67 @@ fn handle_run(
...
+    for (process, cap) in caps_to_remove {
+        println!("terminal: cleaning up process {}, {}", process, cap);
+        Request::new()
+            .target(("our", "kernel", "distro", "sys"))
+            .body(serde_json::to_vec(&kt::KernelCommand::DropCapabilities {
+                target: process,
+                capabilities: vec![kt::de_wit_capability(cap)],
+            })?)
+            .send()?;
+    }
```
dr-frmr: same as above, i'd like to see the number of requests minimized by grouping caps into vectors based on process

kinode/packages/terminal/terminal/src/lib.rs
```diff
@@ -269,6 +334,12 @@ fn handle_run(
             }
         }
     }
+    // always give it the cap to message the terminal back
+    // NOTE a malicious script could use this to drop a ton of caps from other processes
```
dr-frmr: how do we make it so this is not the case?

kinode/src/kernel/mod.rs
```diff
@@ -337,6 +337,33 @@ async fn handle_kernel_request(
...
             let _ = persist_state(&our_name, &send_to_loop, process_map).await;
         }
+        t::KernelCommand::DropCapabilities {
+            target,
+            capabilities,
+        } => {
+            let Some(entry) = process_map.get_mut(&target) else {
+                let _ = send_to_terminal
+                    .send(t::Printout {
+                        verbosity: 0,
```
dr-frmr: verb=1 here please

kinode/src/kernel/mod.rs
```diff
@@ -337,6 +337,33 @@ async fn handle_kernel_request(
...
+                            "kernel: no such process {:?} to DropCapabilities",
+                            target
+                        ),
+                    })
+                    .await;
+                return;
+            };
+            let _ = send_to_terminal
+                .send(t::Printout {
+                    verbosity: 0,
```
dr-frmr: verb=1

kinode/packages/terminal/terminal/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 [dependencies]
 anyhow = "1.0"
 bincode = "1.3.3"
-kinode_process_lib = { git = "https://github.com/kinode-dao/process_lib", tag = "v0.5.9-alpha" }
+kinode_process_lib = { git = "https://github.com/uqbar-dao/process_lib.git", rev = "509b1ed" }
```
dr-frmr: should be `kinode-dao` and tagged 0.6.0 before merge

kinode/packages/terminal/terminal/src/lib.rs
```diff
@@ -217,13 +192,69 @@ fn handle_run(
...
+                        }
+                    }
+                }
+                _ => {
+                    continue;
+                }
+            }
+        }
+    }
+    for (process, cap) in granted_caps.clone().into_iter() {
```
dr-frmr: looks like `.clone()` can be removed here

kinode/packages/terminal/terminal/src/lib.rs
```diff
@@ -396,3 +375,58 @@ fn handle_run(
...
+fn handle_alias_change(
+    state: &mut TerminalState,
+    alias: String,
+    process: Option<ProcessId>,
+) -> anyhow::Result<()> {
+    match process {
+        Some(process) => {
+            // first check to make sure the script is actually a script
+            let Ok(_) = get_entry(&process) else {
+                return Err(anyhow!("terminal: process {} not found", process));
```
dr-frmr, line 389: again, is this check necessary? it's not crazy to have aliases be settable for nonexistent scripts and show a descriptive error when the alias gets used

kinode/src/kernel/mod.rs
```diff
@@ -1092,6 +1143,21 @@ pub async fn kernel(
...
+                        for (grantee, caps) in granter {
+                            let Some(entry) = process_map.get_mut(&grantee) else {
+                                continue;
+                            };
+                            for cap in caps {
+                                entry.capabilities.remove(&cap);
+                            }
+                        }
+                        let _ = responder.send(true);
+                    }
```
dr-frmr, line 1161: need to persist kernel state after this action

kinode/src/main.rs
```diff
@@ -431,7 +431,7 @@ async fn main() {
      */
     let networking_keypair_arc = Arc::new(decoded_keyfile.networking_keypair);
 
-    let (kernel_process_map, db) = state::load_state(
+    let (kernel_process_map, db, reverse_cap_index) = state::load_state(
```
dr-frmr, line 434: solution to this:

since the reverse_cap_index contains zero new data, and is just a re-representation of data in process_map, don't persist it to disk. instead, re-populate the reverse-index on load.

kinode/src/state.rs
```diff
@@ -19,7 +19,7 @@ pub async fn load_state(
     keypair: Arc<signature::Ed25519KeyPair>,
     home_directory_path: String,
     runtime_extensions: Vec<(ProcessId, MessageSender, bool)>,
-) -> Result<(ProcessMap, DB), StateError> {
+) -> Result<(ProcessMap, DB, ReverseCapIndex), StateError> {
```
dr-frmr, line 22: see above: no reason to serialize this data structure and persist!

dr-frmr (changes requested):

kinode/src/main.rs
tadad: yeah that's what I'm doing right?

kinode/src/state.rs
tadad: I'm not! Am I missing something?

kinode/packages/terminal/terminal/src/lib.rs
tadad: I like having the extra check. It saves an extra step from `1. save alias, 2. try using alias and it fails, 3. reset alias` to just `1. get an error for failed alias 2. reset alias`

Plus this check adds basically no time

dr-frmr (approved):

PR #247, opened 2024-02-12
Da/timer

tadad:
Quick PR to formalize the `timer` api

Should be merged with corresponding process_lib changes: https://github.com/kinode-dao/process_lib/pull/54


dr-frmr (approved):
Nice, thanks for this. Do you feel like adding a CancelTimer action? Seems like that should be possible.

PR #254, opened 2024-02-15
IncomingHttpRequest updates

willbach:
## Problem

Currently, it is possible to bind an HTTP path in `http_server` with a URL param like `/test/:test-id`, but it is not possible to get easy access to `:test-id` or the path that was originally bound.

## Solution

This PR forwards the `url_params` and matching `bound_path` which are currently available in, but not being forwarded from, `http_server`.

[Here is the corresponding process_lib update](https://github.com/kinode-dao/process_lib/pull/60)

## Docs Update

[Corresponding docs PR](https://github.com/kinode-dao/kinode-book/pull/135)

## Notes

There will be a repo added that can be used to test http bindings and requests.


Cargo.lock
```diff
@@ -3139,6 +3139,7 @@ dependencies = [
  "open",
  "public-ip",
  "rand 0.8.5",
+ "regex",
```
dr-frmr, line 3142: is this crate still being used? i don't see it in the diff -- if not, please remove from Cargo.lock

Issue #260, opened 2024-02-19
terminal: improve `m` DexX?


nick1udwig:
Currently, using `m` requires an intimate knowledge of the API of whatever process you are interacting with -- and it fails mysteriously if you make a mistake.

This is illustrated by the no-ui file_transfer package. To call it successfully:

```
fake.os > m fake2.os@foo:foo:template.os "ListFiles" -a 5
Mon 2/19 09:14 m: awaiting response for 5s
Mon 2/19 09:14 {"ListFiles":[{"name":"foo:template.os/files/Great Is Thy Faithfulness.m4a","size":3145134}]}
```

```
fake.os > m our@foo:foo:template.os '{"Download": {"name": "Great Is Thy Faithfulness.m4a", "target": "fake2.os@foo:foo:template.os"}}' 
Mon 2/19 09:22 foo worker: begin
Mon 2/19 09:22 file: Great Is Thy Faithfulness.m4a progress: 33%
Mon 2/19 09:22 file: Great Is Thy Faithfulness.m4a progress: 66%
Mon 2/19 09:22 foo worker done: exiting, took 89.831645ms
Mon 2/19 09:22 file: Great Is Thy Faithfulness.m4a progress: 100%
```

So you need to know that you have to await one and not the other. The fail cases look like:

```
fake.os > m fake2.os@foo:foo:template.os "ListFiles"
Mon 2/19 09:37 event loop: don't have 7640537613494437609:terminal:sys amongst registered processes (got message for it from network)
```

```
fake.os > m our@foo:foo:template.os '{"Download": {"name": "Great Is Thy Faithfulness.m4a", "target": "fake2.os@foo:foo:template.os"}}' -a 5
Mon 2/19 09:37 m: awaiting response for 5s
Mon 2/19 09:37 foo worker: begin
Mon 2/19 09:37 file: Great Is Thy Faithfulness.m4a progress: 33%
Mon 2/19 09:37 file: Great Is Thy Faithfulness.m4a progress: 66%
Mon 2/19 09:37 foo worker done: exiting, took 93.304272ms
Mon 2/19 09:37 file: Great Is Thy Faithfulness.m4a progress: 100%
Mon 2/19 09:37 m: SendError: Timeout
```

The second case is more graceful: it does what I want it to, but then prints some error message.

We should either
1. Improve `m` DevX so that "it just works" (preferred, if possible)
[...]

PR #261, opened 2024-02-20
Da/kernel errors

tadad:
## Problem
addresses #260 

## Solution
improved error messages


dr-frmr (approved):
Everything here is great. Just want to add a line in the timeout case suggesting that the caller use `-a`

tadad:
Note I also patched a bug I reported here (#265) - needs a full solution later but this will help a lot in the meantime

PR #264, opened 2024-02-21
fix tests

nick1udwig:
## Problem

`kit run-tests` not working

## Solution

1. Fix tester processes here. 
2. Add a starting "verbosity" level command-line option.
3. Add a "timed out" print.

## Docs Update

[Corresponding docs PR](https://github.com/kinode-dao/kinode-book/pull/140)

## Notes

Solution 3 may be controversial, so please take a look at it.

### DIFF
--- a/kinode/src/kernel/process.rs
+++ b/kinode/src/kernel/process.rs
@@ -351,12 +351,7 @@ impl ProcessState {
             let our = self.metadata.our.clone();
             let timeout_handle = tokio::spawn(async move {
                 tokio::time::sleep(std::time::Duration::from_secs(timeout_secs)).await;
-                print(
-                    &send_to_terminal,
-                    0,
-                    format!("{our}: timed out"),
-                )
-                .await;
+                print(&send_to_terminal, 0, format!("{our}: timed out")).await;
                 let _ = self_sender
                     .send(Err(t::WrappedSendError {
                         id: request_id,
--- a/kinode/src/kernel/process.rs
+++ b/kinode/src/kernel/process.rs
@@ -347,8 +347,16 @@ impl ProcessState {
             let this_blob = blob.clone();
             let self_sender = self.self_sender.clone();
             let original_target = t::Address::de_wit(target.clone());
+            let send_to_terminal = self.send_to_terminal.clone();
+            let our = self.metadata.our.clone();
             let timeout_handle = tokio::spawn(async move {
                 tokio::time::sleep(std::time::Duration::from_secs(timeout_secs)).await;
+                print(
+                    &send_to_terminal,
+                    0,
+                    format!("{our}: timed out"),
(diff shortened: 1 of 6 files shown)


nick1udwig:
To test you must use this PR, [this kit](https://github.com/kinode-dao/kit/pull/91), and [this core_tests](https://github.com/kinode-dao/core_tests/pull/8). You'll have to edit the `core_tests` `tests.toml` to point at your Kinode repo with this branch selected.

nick1udwig:
Some notes:

1. Tests are obliged to wait for a timeout in kns_indexer/app_store before proceeding. It'd be nice if this weren't the case, but I didn't dig into what is going on here. I'd guess the problem is with the nodes not being connected to an ETH RPC.
2. https://github.com/kinode-dao/kit/issues/92
3. https://github.com/kinode-dao/core_tests/issues/7

kinode/src/kernel/process.rs
```diff
@@ -347,8 +347,11 @@ impl ProcessState {
             let this_blob = blob.clone();
             let self_sender = self.self_sender.clone();
             let original_target = t::Address::de_wit(target.clone());
+            let send_to_terminal = self.send_to_terminal.clone();
+            let our = self.metadata.our.clone();
             let timeout_handle = tokio::spawn(async move {
                 tokio::time::sleep(std::time::Duration::from_secs(timeout_secs)).await;
+                print(&send_to_terminal, 0, format!("{our}: timed out")).await;
```
dr-frmr: yeah, controversial indeed.. i don't think this is a good timeout-tracking solution

dr-frmr: is this actually necessary or just a visual debugging tool?

nick1udwig: It is quite hard to debug timeouts without an indication of what is going on. We could increase verbosity level or try to figure out some alternative way to do help devs debug, but the status quo of not informing devs is bad (I spent a day tracking down timeout-related bug because it wasn't clear to me what timeouts were happening, whereas if there were some type of print, it would've been much faster).

dr-frmr: It's hard for me to relate since timeouts *do* show up in the event loop trace, and unless there's a kernel bug in how they get sent back to processes, they should come up in the receive() message queue. In almost all the programs I've written, I have a print that shows any errors that come into my process. Is this an issue that's specific to runtime extension development?

nick1udwig: You are correct. I think maybe this was due to the specific path I followed in bringing `run-tests` into a more usable state or perhaps getting confused about multiple timeouts hitting. Regardless, I've removed the print in f2b8d38

dr-frmr (approved):
Nice, everything looks good now

PR #266, opened 2024-02-22
eth: multi-chain provider model, multi-provider model

dr-frmr:
## Problem

ETH RPC provider runtime module needs to work across all EVM chains and allow for nodes to have multiple providers, of both the RPC and fellow-node variety

## Solution

Rebuilt module to contain chain ID everywhere, and cycle through available providers. Restructure the saved-providers file we hardcode/save to have both RPC URLs and nodes, all labeled by chain ID.

## Docs Update

[Corresponding docs PR](https://github.com/kinode-dao/kinode-book/pull/143)

## Notes

[Corresponding process_lib PR](https://github.com/kinode-dao/process_lib/pull/61)


barraguda (approved):
nice!!

PR #267, opened 2024-02-23
Da/pipe fix

tadad:
Terminal bits and bobs:
- piping was (kind of) broken, fixed it with a docs update as well (https://github.com/kinode-dao/kinode-book/pull/142)
- default pipe time added (5 seconds), e.g. `foo | baz` is equivalent to `foo |5 baz`
- better errors for `hi`

## Notes
- to print anything out in the terminal for a "composable" script, you need to pipe it into `echo`, e.g. `m our@foo:bar:baz "bux" | echo`
  - this is bad devx. I'm open to completely removing pipes since no one is using them anyways


dr-frmr (changes requested):
Piping to echo is no good.. this is not how piping usually works -- if the terminal itself is receiving the response, it should just print it, no? Only if piping is occuring should it skip the print.

tadad:
yeah its tricky because we now drop responses where the request didn't `expects-response`
this means that to get the same functionality, I'd have to do something like `expects-response(60)` to every script that starts in order to maintain that scripts both 1. all scripts are composable via pipes, 2. terminal prints the response if there was no pipe.
This would work but is pretty ugly? Then again piping is also pretty ugly and no one is using it, so a solution could be to just scrap pipes. What do you think?

dr-frmr:
Sounds to me like we need to rethink scripts. The current strategy of treating them as normal programs, but allowing them to behave as functions by "taking arguments" from a request at the beginning of their initialization, then "returning a value" by responding, has run its course and revealed its fundamental limitations.

This is relevant for piping but also script devex as a whole. The current pattern is needlessly verbose given you need to use request/response syntax for what amounts to a function call.

One thing that comes to my mind is adding a script caller function to the wit file, next to `init()` for processes. Could look something like this:
```
script: func(in: tuple<address, request>) -> response
```

dr-frmr:
@tadad I agree that removing pipes may be best for now before we do the work needed to take scripts to the next level.

dr-frmr (approved):

PR #268, opened 2024-02-24
Da/tempfile

tadad:
## Problem
`tempfile` library doesn't work with kinode natively, but does work (in a limited capacity) with wasi

## Solution
edit wasi process config to enable use of `tempfile`

## Notes
I did test this code!


dr-frmr (approved):
We'll need to add capabilities at some point for using this and with what size, but LGTM for now

PR #270, opened 2024-02-29
vfs: sanitize paths

barraguda:
## Problem

Current paths are often manipulated as strings in the vfs, leading way to attacks that use relative paths, like "/mypkg:mypub.os/pkg/../../some_other_pkg/other_file".

## Solution

Canonicalize paths into absolutes, removing any relative paths and checking the absoluts.

## Docs Update

[Corresponding docs PR](https://github.com/kinode-dao/kinode-book/pull/my-pr-number)

## Notes
thanks @bonedaddy for bringing it up in discord!
Caps are matched on literally, so a missing prepended "/" to a path doesn't matter when interacting with the vfs, but when doing get_capability() from the kernel it does.

### DIFF
--- a/kinode/packages/app_store/app_store/src/lib.rs
+++ b/kinode/packages/app_store/app_store/src/lib.rs
@@ -574,7 +574,7 @@ pub fn handle_install(
     state: &mut State,
     package_id: &PackageId,
 ) -> anyhow::Result<()> {
-    let drive_path = format!("/{package_id}/pkg");
+    let drive_path = format!("{package_id}/pkg");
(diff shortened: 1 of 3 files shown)


dr-frmr:
not bad, any changes required in API docs?

barraguda:
> not bad, any changes required in API docs?

nope, not anymore. 

Am doing path normalization instead of canonicalization now as the latter requires the path to exist, which often in our case it does not.

dr-frmr (approved):
LGTM

PR #273, opened 2024-03-05
password work

barraguda:
0.6 auth updates

- hash password UI [x]
- sign hash+timestamp UI [x]
- call appropriate registrar contract for owner [x]
- verify signature is from owner  [x]
- do this for full reset & password reset too [x]
register PR: https://github.com/kinode-dao/register/pull/23


dr-frmr (commented):
Code looks good. Testing UI/UX now...

PR #279, opened 2024-03-16
hotfix: use kit --features flag to pass simulation-mode and fix kit s

nick1udwig:
## Problem

https://github.com/kinode-dao/kit/issues/101

## Solution

@dr-frmr s solution was halfway there. Just needed to propagate the simulation-mode feature flag into the packages. This PR modifies `build.rs` to use the new kit flag which is added in https://github.com/kinode-dao/kit/pull/107

## Docs Update

TODO: update the flags discussion in build & build_and_start.

## Notes

Please take over this PR and the corresponding kit one https://github.com/kinode-dao/kit/pull/107 @dr-frmr 

This is a hotfix and so should be merged directly to main. It bumps to 0.6.1.

Oh, one big note here: there may be weirdness wrt the merge because I branched off of https://github.com/kinode-dao/kinode/commit/61ff56f117a2683099cb946f25355bc3b9b58c40, not HEAD (which is https://github.com/kinode-dao/kinode/commit/fccff76023f2f2a029e54ad06961a628e62ba22d). Please confirm in merge that app_store has the feature flags in it!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2572,7 +2572,7 @@ dependencies = [
 
 [[package]]
 name = "kinode"
-version = "0.6.0"
+version = "0.6.1"
 dependencies = [
  "aes-gcm",
  "alloy-primitives",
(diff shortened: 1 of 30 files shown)


dr-frmr (changes requested):
kns_indexer should also use the flag to stop its indexing work

PR #294, opened 2024-04-02
app store and build fixes redux

nick1udwig:
## Problem

1. Core packages are not rebuilt properly when they are changed,
2. `app_store` does not properly return a failure for certain obvious fail cases.

## Solution

1. Output the required `rerun-if-changed` in `build.rs`,
2. Handle the `Response`s.

## Docs Update

N/A

## Notes

Redo of #291


dr-frmr (approved):
Great, thanks, this is important

PR #295, opened 2024-04-02
Enable sign/verify for apps using node's networking key

dr-frmr:
## Problem

Apps need a way to produce signatures on messages so other nodes can verify their source, outside of the standard networking protocol. Apps must also be able to verify these signatures from other nodes against the PKI

## Solution

The `net:distro:sys` module now has signing and verifying in its core request-response API. Any app with messaging capabilities to `net:distro:sys` can send a `Sign` or `Verify` request with the relevant data and get a response. There is a matching PR in process_lib to expose this functionality cleanly: https://github.com/kinode-dao/process_lib/pull/68

## Docs Update

[Corresponding docs PR](https://github.com/kinode-dao/kinode-book/pull/152)

### DIFF
--- a/kinode/src/net/mod.rs
+++ b/kinode/src/net/mod.rs
@@ -982,30 +982,13 @@ async fn handle_local_message(
         parse_hello_message(our, &km, body, kernel_message_tx, print_tx).await?;
         Ok(())
     } else {
-        let maybe_response = match rmp_serde::from_slice::<NetAction>(body)? {
-            NetAction::ConnectionRequest(_) => {
-                // we shouldn't receive these from ourselves.
-                None
-            }
-            NetAction::KnsUpdate(log) => {
-                pki.insert(
-                    log.name.clone(),
-                    Identity {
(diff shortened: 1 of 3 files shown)


nick1udwig:
Does it make sense to have signing for arbitrary payloads? I.e. I call `sign(data: &[u8])` and get back the sig of those bytes?

dr-frmr:
> Does it make sense to have signing for arbitrary payloads? I.e. I call `sign(data: &[u8])` and get back the sig of those bytes?

What we have now is signing of arbitrary payload BUT the process ID is prefixed. This is done for security reasons: if an app was able to produce signatures using the node's networking key, it could spoof signatures used by any other app for any purpose. This way, signed payloads are always irrevocably linked to the specific app which produces it.

...if we can create a use case for signing arbitrary payloads, we can add that to the API in the future.

nick1udwig:
Gotcha, I was misreading the code. Thanks for clarifying!

PR #296, opened 2024-04-02
fix login page and add compiled types to register-ui

0x70b1a5:


dr-frmr (approved):

dr-frmr (changes requested):
remove ts autogen files

dr-frmr (approved):

PR #299, opened 2024-04-04
Dockerize Kinode

0xlynett:
## Problem
There is no Docker image for Kinode.

## Solution
I made a Docker image for Kinode.

## Docs Update
https://github.com/kinode-dao/kinode-book/pull/156

## Notes

I couldn't get it working on Alpine, so the final image is a bit large (Debian 12 Slim)

Docker Hub: https://hub.docker.com/r/0xlynett/kinode

TODO:
- use a different account other than my personal one as the namespace
- set up a workflow to automatically build new releases
- use Alpine if possible


0xlynett:
Added the docs update PR

nick1udwig:
Tested and it works great, thank you very much for this PR!

One thing I found a little confusing (not reading the docs, just looking at code/playing with cmdline) was that the home directory lives in `/kinode` and the binary lives in `/bin/kinode`. Does it make more sense if we call the home dir something like `/home` or `/kinode-home`?

0xlynett:
yeah thats a good idea, i'll change the dockerfile to use `/kinode-home`
not sure how it would interact with standard FHS folders like `/home`

0xlynett:
https://github.com/0xlynett/kinode-book/commit/22cb3332b7b917f360990356310838524f45a4a3

nick1udwig (approved):
Tested on Linux & macOS and it works great on both. Nice work, thanks for the contribution :+1:

dr-frmr (approved):

PR #305, opened 2024-04-15
Da/unpublish

tadad:
## Problem
app_store UI did not have an unpublish feature

## Solution
Now it appears in a list of packages that your current connected wallet owns

NOTE: we could reorganize this screen into two new ones: "publish new package" and "manage existing packages". I think that's coming eventually but maybe not today

Screenshot 2024-04-15 at 5 51 19 PM

### DIFF
--- a/kinode/packages/app_store/ui/src/App.tsx
+++ b/kinode/packages/app_store/ui/src/App.tsx
@@ -63,6 +63,7 @@ function App() {
 
   useEffect(() => {
     provider?.getNetwork().then(network => {
+      // TODO: don't have this fire 20 times on load
       if (network.chainId === ChainId.OPTIMISM) {
         setPackageAbi(PackageStore__factory.connect(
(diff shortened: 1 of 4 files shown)


kinode/packages/app_store/ui/src/App.tsx
```diff
@@ -63,6 +63,7 @@ function App() {
 
   useEffect(() => {
     provider?.getNetwork().then(network => {
+      // TODO: don't have this fire 20 times on load
```
dr-frmr, line 66: I really really wish we could fix this for 0.6.3

dr-frmr (approved):
Looks fantastic. Just wish we could get the overabundance of RPC calls fixed

kinode/packages/app_store/ui/src/App.tsx
tadad: yeah if you want me on that, I can make another branch and try to get it done for 0.6.3

PR #308, opened 2024-04-16
hotfix: app_store rpc

barraguda:
## Problem

Initializing the contract abi instance in the root app.tsx leads to trying a preconfigured RPC_URL if wallet is not connected, resulting in console errors

## Solution

Make provider and abi optional props, only passed to the publish component.

## Notes

We should review some of our core UI wallet interactions


tadad (approved):
Looks good. I noticed some of these oddities with the code when doing the unpublish PR - didn't realize they were causing the RPC issues

PR #316, opened 2024-04-25
Kernel: process restart even on crash

dr-frmr:
## Problem

We've gone back and forth on this, but the kernel currently cancels restarts for processes that crash. This turns out to be harmful to the overall application model, because it complicates the story for managing `OnExit` behavior and has the effect of making process crashes totally verboten. Processes need to be allowed to panic and recover, and Restart is our primary recovery mechanism. There are alternatives as described [here](https://gist.github.com/dr-frmr/bd2a0dd299a083034c0ba79d46ba2899) but they are much more involved.

## Solution

Remove the carve-out for crashes and honor `OnExit` on crash.

## Docs Update

N/A but we SHOULD document this in book somewhere: https://gist.github.com/dr-frmr/bd2a0dd299a083034c0ba79d46ba2899

### DIFF
--- a/kinode/src/kernel/process.rs
+++ b/kinode/src/kernel/process.rs
@@ -713,61 +713,61 @@ pub async fn make_process_loop(
                     lazy_load_blob: None,
                 })
                 .await?;
-                let _ = send_to_terminal
-                    .send(t::Printout {
-                        verbosity: 1,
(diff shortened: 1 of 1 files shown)


nick1udwig (approved):
lgtm :rocket:

PR #317, opened 2024-04-25
Update Dockerfile to download from GitHub releases

0xlynett:
## Problem

It takes a lot of time and wasted compute to rebuild the Docker image for every new Kinode update, when prebuilt binaries exist.

## Solution

This change downloads the binary directly from the latest GitHub release rather than compiling every single time.

## Docs Update

N/A

## Notes

Ideally in the future we'd want to decentralize more of the updates and code stuff.

version 0.7.0 is on docker hub now btw

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,30 +1,20 @@
-# syntax=docker/dockerfile:1
-FROM rust AS builder
+FROM debian:12-slim AS downloader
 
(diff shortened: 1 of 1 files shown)


nick1udwig (approved):
LGTM; tested & it works. Thanks for the change to make maintenance easier! :rocket:

PR #320, opened 2024-04-29
app store: add rebuild-index endpoint for frontend

dr-frmr:
## Problem

Users may need to easily refresh their app store's representation of the onchain listings. This feature was available via CLI but now can be triggered from frontend.

## Solution

Adds POST endpoint `/apps/rebuild-index` to `main:app_store:sys`

Adds refresh button to app store listings page

## Docs Update

N/A


dr-frmr:
All done!

PR #328, opened 2024-05-03
Widgets

0x70b1a5:
## Problem

- We don't have widgets.
- We don't see all our installed packages on the home screen.
- We don't have a Launch button in the appstore for launching a package that we just installed.

## Solution

- Add widgets.
- Show installed packages on the home screen.
- Add a Launch button to the appstore.

## Docs Update

We don't mention in the Book that apps have to message homepage:sys in order to show up there - I'm not sure that writing a whole page about this is in-scope, especially given that we need this out ASAP.


0x70b1a5:
image

kinode/packages/homepage/homepage/src/lib.rs
```diff
@@ -21,6 +21,7 @@ enum HomepageRequest {
         label: String,
         icon: String,
         path: String,
+        widget: Option<String>,
```
dr-frmr, line 24: This is great, and will be backwards compatible as an optional field.

kinode/packages/homepage/homepage/src/lib.rs
```diff
@@ -98,42 +116,70 @@ fn init(our: Address) {
...
+                                        None,
+                                        format!("Missing id").into_bytes(),
+                                    );
+                                    println!("failed to parse app id");
+                                    for key in url_params {
+                                        println!("{} {}", key.0, key.1);
+                                    }
+                                    println!("All application IDs currently stored:");
+                                    for key in app_data.keys() {
+                                        println!("{}", key);
```
dr-frmr: Please remove the prints here for final version

kinode/packages/homepage/homepage/src/lib.rs
```diff
@@ -98,42 +116,70 @@ fn init(our: Address) {
...
+                                    }
+                                    continue;
+                                };
+                                let app = app_data.get_mut(&app_id.to_string());
+                                if let Some(app) = app {
+                                    app.is_favorite = !app.is_favorite;
```
dr-frmr: Does this edit of an app's favorite status get persisted? If I set favorites, then reboot node, are they still there?

dr-frmr: Ok, I tested, and looks like the answer is no. You'd need to add logic here that persists the homepage state between boots, if we are to put 'favoriting' logic here.

To me, this is an argument for keeping 'favorites' logic on the frontend (using browser localstorage I suppose), because `homepage` is currently stateless and keeping it that way is nice.

dr-frmr (changes requested):
Left some comments -- code looks good overall.

The frontend has some minor issues to address -- as described in comment, I think 'favoriting' is better suited for browser localstorage as it turns out, given that it's not persisted between boots currently and doing so would involve turning homepage into a stateful app.

Style-wise, I think the widgets section simultaneously does not need a "Widgets" title/label, and needs to be clearer about what the widgets are. I think a good way to do this would be to make the maximum widget size smaller such that it's not taking up the full horizontal space, and add a label to each individual widget that notes which app it's from.

After these changes, let's get another review from @jtneev on pure design.

Edit: Oh also, I think we need to make it clear somehow that the "default" apps that get shown on homepage despite having zero frontend functionality are there just for... show? They should be gray or something and unclickable.

kinode/packages/homepage/homepage/src/lib.rs
0x70b1a5: No problem.

0x70b1a5:
- [x] Save app favorites in localStorage instead of sessionStorage
- [x] Make widgets smaller and change labelling
- [x] Disable mouse behavior for un-UI-ful apps

dr-frmr (changes requested):
Screenshot 2024-05-06 at 2 27 57 PM

Style review:
- Icons need to be smaller, and if/when they overflow, needs to look a bit better -- cutoffs are choppy right now.
- Favorited apps still appearing in main list is confusing
- The two gear icons at different places on the right edge of the page are confusing. It's also weird that the settings page has a big icon in the apps list and also has a link in the corner. I propose we remove the system settings gear in top-right and use *that* location for the widgets settings modal.
- Widgets should be "in front of" apps that are downloaded but not favorited. Apps that are downloaded but not favorited need to be de-emphasized in general as this section will grow very large as a user downloads things.

hyper-edgar:
A few more comments: 

- Let's change the Settings app to say "System Settings" and change the icon so it's not a gear (or change the widget settings gear). We can't have two gears that mean different things.
- Spacing is off in this picture. It's crowded. Either we need something that adjusts for screen size (best if you can do it), or we need to develop for default mac laptop size. 
- New order is this: Favorited Apps, Widgets, and below a collapsible "All Installed Apps" section. Favorited apps can stay there (updated from what Ben said above). 
- In the App Store, all apps should have a default icon like the one on the homepage if they don't specify an image

This is very close, nice work

0x70b1a5:
@jtneev @dr-frmr 
- [x] Appstore icons will be color dots when image is unset
- [x] Components are generally smaller and look better on default screens
- [x] "All apps" really does show all apps, and it expands upwards from the bottom in a drawer that is hidden by default
- [x] Widget settings button in upper right, using a wrench-and-screwdriver icon instead of gear
- [x] Widgets are now above "All Apps"

dr-frmr (approved):
Awesome, looks good to go.

PR #334, opened 2024-05-06
fix build.rs `src/` behavior

nick1udwig:
## Problem

We don't rebuild core processes when their src/ is changed.

## Solution

When we are inside a directory that is contained in `rerun_files`, print all entries therein.

## Docs Update

N/A

## Notes

None

### DIFF
--- a/kinode/build.rs
+++ b/kinode/build.rs
@@ -24,17 +24,28 @@ fn get_features() -> String {
 }
 
 fn output_reruns(dir: &Path, rerun_files: &HashSet<String>) {
-    if let Ok(entries) = fs::read_dir(dir) {
-        for entry in entries.filter_map(|e| e.ok()) {
-            let path = entry.path();
-            if path.is_dir() {
-                // If the entry is a directory, recursively walk it
-                output_reruns(&path, rerun_files);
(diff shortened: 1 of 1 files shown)


dr-frmr:
@nick1udwig is there an amortizeable performance cost to all this..? or is it negligible/cost of doing business

nick1udwig:
So the "correct" behavior is that on a rebuild we go and try to `kit build` all the `packages/`. I THINK this PR should do that, but I've thought like 5 iterations of `build.rs` should do that so...

In any case, `kit build` does a good job of rebuilding only when necessary. You can experience this yourself when you `kit b` a package: first time build takes a while; second time with no changes is nearly instant; third time with changes is usually 1-2 seconds on my machine.

dr-frmr:
Oh no I just meant performance of traversing all the directories to find changes and cargo-print them. I suppose that's meaningless compared to build time itself.

dr-frmr (approved):
LGTM

nick1udwig:
> Oh no I just meant performance of traversing all the directories to find changes and cargo-print them. I suppose that's meaningless compared to build time itself.

Ah, I see. Yeah, I think you are correct: small wrt overall build time.

Issue #329, opened 2024-05-06
bug: import keyfile fails 


0xlynett:
**Describe the bug**
when trying to import a keyfile, it errors with `Request body deserialize error: missing field password_hash at line 1 column 602`. this is likely due to a field being changed on the frontend but not the validation endpoint

**To Reproduce**
1. Import keyfile
2. Fail

**Expected behavior**
Importing the keyfile successfully.

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser Chrome
 - Kinode version 0.6.2 (also happens on later versions)

**Additional context**

request body looks like:
```
payload: <the keyfile here>
password: <the password hash here>
```
but it expects `password_hash` instead of `password`

dr-frmr:
Thanks for this report. Definitely an oversight when we changed the field name a couple versions ago. Will be fixed in next release

PR #343, opened 2024-05-14
fix import keyfile

0x70b1a5:
## Problem

Importing a keyfile doesn't work

## Solution

Work it

## Testing

- Import a keyfile to a fresh boot and verify normal startup


0xlynett:
addresses https://github.com/kinode-dao/kinode/issues/329

dr-frmr (approved):

PR #347, opened 2024-05-16
develop 0.8.0

nick1udwig:
- add WIT APIs
- add API-fetching interface to app store package
- upgrade `kinode.wit` to 0.8. only changes: add `package-id` and add `target` to `send-error`.
- rewrite app store for clarity and extensibility
    - get auto update working
    - TODO: switch to better file_transfer worker implementation
- rewrite networking and add TCP networking. changes to kernel and core types where necessary to make networking protocol clean to be implemented across many transport protocols

- get secure subdomains working 100%
    - decided to delay switching key apps to it for future release

- general cleanups and bug-fixing
- published process_lib 0.8.0 and switched all packages to using that by tag


kinode/src/state.rs
```diff
@@ -404,7 +404,7 @@ async fn bootstrap(
         let our_drive_name = [package_name, package_publisher].join(":");
         let pkg_path = format!("{}/vfs/{}/pkg", &home_directory_path, &our_drive_name);
         // delete anything currently residing in the pkg folder
-        fs::remove_dir_all(&pkg_path).await?;
+        let _ = fs::remove_dir_all(&pkg_path).await;
```
nick1udwig: I'd recommend doing the following instead. Swallowing `Result`s can lead to stuff that is hard to debug
```
let pkg_path = std::path::PathBuf::from(pkg_path);
if pkg_path.exists() {
    fs::remove_dir_all(&pkg_path).await?;
}
```

dr-frmr (approved):

PR #349, opened 2024-05-16
native-tls so that wss:// type urls will work

Tenari:
## Problem

couldn't connect to holium's staging comfyui server via websockets lib because it needed wss:// instead of ws:// (tls)

## Solution

include tls in the binary
```

### DIFF
--- a/kinode/src/http/client.rs
+++ b/kinode/src/http/client.rs
@@ -218,7 +218,8 @@ async fn connect_websocket(
     // Connect the WebSocket
     let ws_stream = match connect_async(req).await {
         Ok((ws_stream, _)) => ws_stream,
-        Err(_) => {
+        Err(e) => {
+            println!("{e:?}");
             return Err(HttpClientError::WsOpenFailed {
                 url: url.to_string(),
             });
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: kinode/src/http/client.rs [hunk 1]
```diff
@@ -218,7 +218,8 @@ async fn connect_websocket(
     // Connect the WebSocket
     let ws_stream = match connect_async(req).await {
         Ok((ws_stream, _)) => ws_stream,
-        Err(_) => {
+        Err(e) => {
+            println!("{e:?}");
```
dr-frmr: Please remove this!

dr-frmr (commented):
Remove the print and good to go! Can also replace the print with an official one using the Sender for terminal, probably good to hide it at debug level = 1

Tenari:
@dr-frmr is this the correct way to print now? https://github.com/kinode-dao/kinode/pull/349/files#diff-3d22168e645c4d6513c023acf9ed57ed3d5950d1b0ae3f5ba48644c8bf2457d9R222

dr-frmr (approved):
Yes it is! LGTM

PR #350, opened 2024-05-17
vfs: remove unused select

nick1udwig:
## Problem

We have a `select!` where there is only one item: must be a leftover from when there was >1 item.

## Solution

Remove the `select!`

## Testing

Build it

## Docs Update

None

## Notes

None


dr-frmr:
lgtm

barraguda (approved):
👍

PR #352, opened 2024-05-17
appstore widget links

0x70b1a5:
## Problem

appstore widget doesn't link to app page

## Solution

add links

## Testing

Click on an app in the widget and verify it opens a new tab with the app page in it.

## Notes

I tried to use `window.postMessage` with the iframe to open the link in the same tab, but for whatever reason it wasn't taking. Investigation needed.


dr-frmr (approved):
LGTM

PR #379, opened 2024-06-05
Kinomap

barraguda:
WIP.

notes:
- current setting of ip, port and pubkey takes 3 txs. Quick on local, but will be a single multicall once we get it into the contracts. [x]
- no sending KnsUpdates to net:sys yet, need to determine when we've indexed enough logs to determine a node "complete" to send to net [x]
- separate core process that indexes all notes and their values? (I suspect a better approach would be a process that has good library functions, and can index/subscribe on demand instead)
- currently registering a name will not include it's .dev/.os top level zone. This is due to how kit loads state from json, and it doesn't include events (that the kns_indexer should pick up). This is an open anvil [issue](https://github.com/foundry-rs/foundry/issues/7502) that I might take a look at to fix, or we have to reconsider fakenet approach [x]


kinode/packages/app_store/pkg/ui/assets/Arc-R3PUWRPJ-SiUbLWTa.js
dr-frmr, line 1: all of these individual js asset files really really need to be compressed if you're going to commit them. we'll need to squash merge this PR

kinode/packages/app_store/app_store/src/utils.rs
```diff
@@ -74,10 +76,13 @@ pub fn fetch_state(our: Address, provider: eth::Provider) -> State {
 }
 
 pub fn app_store_filter(state: &State) -> eth::Filter {
+    let notes = vec![keccak256("~metadata-uri"), keccak256("~metadata-hash")];
+
     eth::Filter::new()
         .address(eth::Address::from_str(&state.contract_address).unwrap())
         .from_block(state.last_saved_block)
         .events(EVENTS)
+        .topic3(notes)
```
dr-frmr, line 85: this is so cool

kinode/packages/app_store/pkg/ui/assets/Arc-R3PUWRPJ-SiUbLWTa.js
barraguda: yes, same case for register_ui ones.. there are too many anyway, looking into compressing soon

dr-frmr (approved):
LGTM

PR #380, opened 2024-06-05
fix: vfs crash due to ProcessIdParseError display function

dr-frmr:
## Problem

Fix a bug in ProcessIdParseError display logic that was causing VFS crashes


barraguda (approved):
lgtm, let's rethink the queue and spawning in another PR

PR #383, opened 2024-06-06
eth: explicit rpc error return

barraguda:
## Problem

If I send a signed transaction that will fail on the rpc level, for example with too low a gas level, the eth runtime currently does a few things:

receives error, assumes something is wrong with the provider, returns a NoRpcFound error, restart subscriptions.

## Solution

If ErrResponse from provider, to_string() it and return it to the user so they know what went wrong.

## Notes

corresponding process_lib update: https://github.com/kinode-dao/process_lib/pull/80


dr-frmr (approved):
Perfect

PR #404, opened 2024-06-17
reloadable widgets

0x70b1a5:
## Problem

On the mobile app, especially while we're still in alpha/beta, a user might get a widget into an unrecoverable state (navigating via a link to a page inside the iframe, for example).

## Solution

In our iframes, listen for messages from the parent window. On `reload` message, reload the iframe.

## Testing

- [ ] Click the reload button in the ios app and verify that a widget in a broken state returns to functional state.

## Docs Update

WIP. I intend to add a Widgets page to the docs with this PR, but I first need to verify that the reload message works as intended.

## Notes

10/10, no notes


0x70b1a5:
Found a way.

Issue #303, opened 2024-04-12
bug: http_client WS breaks Kinode Req/Res


nick1udwig:
**Describe the bug**
Client WS Pushes break Kinode Request/Response mental model just like how Server WS pushes used to (resolved in #218)

**To Reproduce**
1. Receive a Request (id `123`) you want to eventually Respond to
2. Do some work involving WS client, receiving pushes (which are always Requests, and hence, reset the `prompting_message`)
3. Try to Respond

Instead of routing to `123`, it will route to http_client.

**Expected behavior**
Route to `123`.

**The fix**

Requires some thought. In #218 it was assumed we'd always have a Req/Res happening. In my particular usecase, I am connecting to a WS, receiving some arbitrary number of pushes, and then closing the connection.

One hack we can do is when connecting to the WS, you can set if you want to be receiving Requests or Responses from the server. This fixes my case but not the general case.

nick1udwig:
This makes me think the #218 solution may not be sufficiently general either.

Issue #219, opened 2024-01-30
Design feedback on lazy_load_blob


manuel:
I think the idea of enabling middleware to just receive the head of a request, and forward it, and only have the final recipient return a response [(link)](https://book.kinode.org/processes.html#inheriting-a-lazy_load_blob) is a great idea (out of curiosity, did you come up with that or did you get it from somewhere else?)

However, I see an issue with the current implementation: AFAICT, only the last lazy_load_blob received by a process is remembered by the kernel. So what if the middleware needs to make another request before forwarding that lazy_load_blob to the next destination, e.g. to fetch some info?

dr-frmr:
Thank you for the feedback! I'm sure others have put together similar structures, however we were inspired to add this functionality purely via benchmarking various Wasm experiments and observing the bottleneck often being moving bytes into a process.

I think you're right that the current implementation is a bit unfinished. We've got a lot of work left to do before `1.0`...

Looking at the kernel logic around process state, one can see we have a `prompting_message` and `last_blob`. Something I have been very interested in doing is restructuring the `contexts` map, which keeps track of outstanding requests, to also manage these two fields. I think it could be very intuitive to have a nesting logic where total context (including the `lazy_load_blob`) is popped off a stack as you receive responses.

The issue I'm having so far is that responses themselves can also contain blobs, of course, and I haven't figured out a good way for the programmer to "traverse" a "stack" of blobs/`prompting_message`s.

manuel:
Each request could have a unique ID, and you could fetch a particular lazy_load_blob via that ID?

dr-frmr:
We have unique IDs for every message, but currently don't expose that to the process. The thought is that processes should use state of some kind to manage their data rather than keep track of a bunch of request IDs. I'm hesitant to change this property because it's been nice, so far, to treat messages as nothing more, and build state machines that react to them.

With regards to fetching, we've so far been able to avoid the issue of memory leakage by *only* tracking outgoing requests that expect a response (which *must* be associated with a timeout). If we added functionality to retrieve arbitrary data from past messages, wouldn't we need to start baking in some logic around expiry, which becomes difficult to mentally model as a process developer?

dr-frmr:
> So what if the middleware needs to make another request before forwarding that lazy_load_blob to the next destination, e.g. to fetch some info?

[...]

PR #411, opened 2024-06-18
kernel: only overwrite request prompting_message when there could be a reply to it

nick1udwig:
## Problem

Untitled

## Solution

Only update the `prompting_message` when new one could conceivably be useful.

## Testing

Ask me

## Docs Update

TODO

## Notes

This is a small change but should be carefully considered because
1. It is technically breaking,
2. Even if we decide we want it, we may want to do a similar thing with the `Response` `None` context case, in the match case just below.

Let's discuss @dr-frmr

### DIFF
--- a/kinode/src/kernel/standard_host.rs
+++ b/kinode/src/kernel/standard_host.rs
@@ -101,7 +101,10 @@ impl process::ProcessState {
     ) -> Result<(wit::Address, wit::Message), (wit::SendError, Option<wit::Context>)> {
         let (mut km, context) = match incoming {
             Ok(mut km) => match km.message {
-                t::Message::Request(t::Request { ref expects_response, .. }) => {
+                t::Message::Request(t::Request {
+                    ref expects_response,
+                    ..
+                }) => {
                     self.last_blob = km.lazy_load_blob;
                     km.lazy_load_blob = None;
                     if expects_response.is_some() || km.rsvp.is_some() {
--- a/kinode/src/kernel/standard_host.rs
+++ b/kinode/src/kernel/standard_host.rs
@@ -101,10 +101,13 @@ impl process::ProcessState {
     ) -> Result<(wit::Address, wit::Message), (wit::SendError, Option<wit::Context>)> {
         let (mut km, context) = match incoming {
             Ok(mut km) => match km.message {
-                t::Message::Request(_) => {
+                t::Message::Request(t::Request { ref expects_response, .. }) => {
                     self.last_blob = km.lazy_load_blob;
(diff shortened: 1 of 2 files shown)


nick1udwig:
Hmm, no I think this change as-is might be wrong. The problem is that the state can become fractured, e.g.:

https://github.com/kinode-dao/kinode/blob/88db6ef8e3c1b2b0b8b8869fe22eb77bf6f04a97/kinode/src/kernel/standard_host.rs#L108-L113

You can end up in a situation where the `prompting_message` remains unchanged but the blob gets updated. This is strictly wrong behavior.

nick1udwig:
Well, I'm not certain that these two fields need to track 1-1, but it might lead to hard to debug situations. Need to think.

The way that this could be OK is that usually you only care about one thing at a given time: either you want to blob or you want to Respond. Sometimes you want both. In each of these cases you are OK, as written, since the blob is there when you expect it to be and the Response is routed to the most recent place you could possibly Respond to.

nick1udwig:
I think the main question becomes:

Is this mental model too hard to work with?

Existing mental model: I can get blob of & reply to most recent message (even if that message does not expect response/have RSVP).

New mental model: I can get blob of most recent message. I can reply to most recent message that expects response/has RSVP.

In first case you only need to know whether most recent message expects response/has RSVP.

In the case in this PR you need to know expects response/has RSVP for potentially many messages -- i.e. if you have N-1 most recent messages with not expects response/RSVP and then the Nth does er/RSVP, then the Nth is the one that will receive the response.

In one sense this PR is just enforcing expects response/RSVP to actually mean something from the side of the sender: you cannot reply to a message that does not ER/have RSVP.

nick1udwig:
Aha, this is just a resurfacing of an old idea https://github.com/kinode-dao/kinode/issues/219#issuecomment-1917694538

nick1udwig:
Yeah, so that comment on #219 is when I was thinking about Kinode extensions. You can see the extension PR is right around the same time as that comment. And indeed, this change means that you can receive WS pushes without messing up your Req/Res context (since they are always not ER and no RSVP). This is really good! Also solves the similar #303.

HOWEVER: is it too much to ask of devs to understand the ER/RSVP of the Requests that come in?

Effectively this introduces a new type of Request to our system, the "one-off" Request. It does not start a new chain. It is just a packet of information disconnected from any other Req/Res chain and not interrupting current chains.

nick1udwig:
shapes at 24-06-18 10 25 26

alpha & beta here work on master. gamma & delta only work with this PR

dr-frmr (approved):

PR #414, opened 2024-06-19
notify

0x70b1a5:
## Problem

- Notify app needs to exist even a little bit so people can play around with it
- App store needs mobile UI tune up
- Homepage should confirm with user if they want to hide a widget

## Solution

- [x] Make notify app with widget
- [x] Fix up appstore mobile UI
- [x] Add confirmation dialogue for widget hide on homepage

## Testing

- [x] Verify appstore looks alright on mobile screen sizes
- [x] Verify widgets ask before hiding
- [x] Add notifications to notify and watch them show up in the widget

## Notes

This is very much a v0.1 for notify, but I think "push things to a widget" is a suitable v0.1 if we're being honest. The next feature is silencing processes. Actual device push notifs will be blocked until the mobile app is approved for TestFlight.

Also, I found some merge conflicts in the Settings and Chess `Cargo.lock` files that had been committed at some point in the past. I took the more recent of each version where necessary, but would have preferred to regenerate them (deleting them and then `kit b` and/or rebuilding binary didn't cut it).


kinode/packages/app_store/ui/src/pages/StorePage.tsx
```diff
@@ -97,12 +100,18 @@ export default function StorePage() {
   );
 
   const tryRebuildIndex = useCallback(async () => {
+    if (!window.confirm('Are you sure you want to rebuild the app index? This may take a few seconds.')) {
```
dr-frmr, line 103: this is good

kinode/packages/homepage/ui/src/store/persistentStore.ts
```diff
@@ -34,6 +34,9 @@ const usePersistentStore = create<PersistentStore>()(
       setFavoriteApps: (favoriteApps: PersistentStore['favoriteApps']) => set({ favoriteApps }),
       toggleWidgetVisibility: (package_name: string) => {
         const { widgetSettings } = get()
+        if (!window.confirm(`Really hide this widget?`)) {
```
dr-frmr, line 37: i really dislike this alert box. IMO it needs to either be an in-page confirmation or have no confirmation at all. i'm not sure why this action needs to be confirmed, given how straightforward it is to re-toggle the widget on.

another option would be to completely remove the "hide widget" button from the home screen and only allow that operation in the settings modal. that would be my preference.

kinode/packages/notify/notify/src/lib.rs
```diff
@@ -0,0 +1,266 @@
+#![feature(let_chains)]
+use std::collections::HashMap;
+
```
dr-frmr, line 3: remove blank line in between `use` statements please

kinode/packages/notify/pkg/manifest.json
```diff
@@ -0,0 +1,16 @@
...
+        "on_exit": "Restart",
+        "request_networking": false,
+        "request_capabilities": [
+            "http_server:distro:sys",
+            "homepage:homepage:sys"
+        ],
+        "grant_capabilities": [
+            "http_server:distro:sys"
+        ],
+        "public": true
```
dr-frmr, line 14: should the notify app really be public? shouldn't other apps need to request messaging capabilities in order to push notifications? wondering what the overall security model for this program should look like

0x70b1a5:
Appstore fixes moved to #418

Issue #416, opened 2024-06-19
bug: system doesnt recognize caps after a bad terminal command


jurij-jukic:
**Describe the bug**

- a bunch of messages of this kind start popping up for various permutations of processes:
```
Wed 15:53 event loop: process http_server:distro:sys doesn't have capability to message process main:command_center:appattacc.os
```

- my guess is that this starts happening after running a command which inputs the wrong type. i.e. worker_address here is supposed to be an Address, but i am inputting a string:
```
m astronaut.os@main:command_center:appattacc.os '{"BackupRetrieve": {"node_id": "sour-cabbage.os", "worker_address":""}}'
```

so you can probably recreate this with any app you are working by simply inputting a wrong type. but maybe i'm wrong and if something else is the cause, here is how to reproduce.

**To Reproduce**
1. install the following commit of [command center](https://github.com/kinode-dao/command_center/pull/1/commits/5fe9847ea7e993c12a61d3307f0b2647c5458995
) on your kinode 
2. run the command:
```
m astronaut.os@main:command_center:appattacc.os '{"BackupRetrieve": {"node_id": "sour-cabbage.os", "worker_address":""}}'
```
3. a bunch of previously successful caps will stop working

nick1udwig:
https://discord.com/channels/1186394868336038080/1186421052071477428/1253186534266507367
https://discord.com/channels/1186394868336038080/1186421052071477428/1253187341045072044

dr-frmr:
Working on this

dr-frmr:
@nick1udwig @jurij-jukic check out #427

nick1udwig:
#427 fixed this

PR #427, opened 2024-07-02
fix revoke all behavior, make caps printing on verbosity=3 cleaner

dr-frmr:
## Problem

Issue #416 

## Solution

It seems like revoking all capabilities on process spindown wasn't working for a while, which led to us not observing this issue. Now that RevokeAll is operating as expected, removing capabilities generated by a process when it gets killed or crashes, we need to re-grant capabilities in the OnExit::Restart case.

I came up with a simpler solution: adding a flag to the kill procedure that simply skips over the RevokeAll call. Then I use that flag in the Restart case on process spindown.

## Testing

See #416 

## Docs Update

We should thoroughly document this whole thing -- I'll get on that

## Notes

Also made capabilities print more nicely in terminal on verbosity 3.

### DIFF
--- a/kinode/src/kernel/mod.rs
+++ b/kinode/src/kernel/mod.rs
@@ -306,7 +306,9 @@ async fn handle_kernel_request(
                 .expect("event loop: fatal: sender died");
             None
         }
+        //
         // send 'run' message to a process that's already been initialized
+        //
         t::KernelCommand::RunProcess(process_id) => {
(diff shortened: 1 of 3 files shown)


nick1udwig (approved):
Works as advertised :+1: 

Interesting use of `metadata` field. I don't know that I feel great about it, but it is backwards compatible

Issue #425, opened 2024-07-02
bug: navigating to app before logging in on fake node never allows login


nick1udwig:
**Describe the bug**
If you navigate to a served endpoint before logging in to a node, you get some logic page talking about secure subdomain stuff that always redirects to itself rather than logging you in.

**To Reproduce**
```
kit f

# new terminal
kit n foo --ui
kit bs foo
```
Navigate to `https://localhost:8080/foo:foo:template.os`

**Expected behavior**
1. There should be no mention of secure subdomains.
2. We should be able to login when navigating directly to app & be redirected properly.

**Additional context**
https://discord.com/channels/1186394868336038080/1186421052071477428/1257490840729813105

This is a problem for both real and fakenodes. Its somewhat harder to experience on realnodes because when you run them yourself
1. You are automatically sent to the login page,
2. Apps are not yet served until you login.

The way you can experience this on a realnode is with hosting. E.g., https://templater.hosting.kinode.net/chess:chess:sys

nick1udwig:
Fixed by #428

PR #428, opened 2024-07-02
fix: if hitting path before login, redirect to login

dr-frmr:
## Problem

#425 

## Solution

If a path is not bound to a subdomain, instead of directly serving subdomain-focused login page, redirect to `/`.

The only problem with this is that the user then has to navigate back to the path they were trying to hit directly pre-log-in.

## Testing

See #425 

## Docs Update

N/A

### DIFF
--- a/kinode/src/http/server.rs
+++ b/kinode/src/http/server.rs
@@ -218,7 +218,6 @@ pub async fn http_server(
     ));
 
     while let Some(km) = recv_in_server.recv().await {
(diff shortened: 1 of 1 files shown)


nick1udwig (approved):
Tested: fixes the bug. Thanks for this! I've opened #429 to track the slight UX improvement, but this PR is far better than the current state, so we should merge as-is.

PR #434, opened 2024-07-08
fix caps

nick1udwig:
## Problem

Caps don't work properly. This is because of the `contains_key()` call here: https://github.com/kinode-dao/kinode/blob/main/kinode/src/kernel/mod.rs#L990

## Solution

`contains_key()` wants a `PartialEq` as well as a `Hash`. `PartialEq` JSON-ifies the `params` field before comparison, but `Hash` does not. Make `Hash` JSON-ify `params`.

## Testing

In the `kinode-book/src/code`, you can `kit t` on all but the `http` packages and it will fail on main, but work on this branch.

## Docs Update

N/A

## Notes

I also merged the hotfix commits here. The actual fix is commit 7413646


dr-frmr (approved):
Great catch, we should merge this and put out 0.8.3 right away. Surprised it didn't break more stuff!

PR #445, opened 2024-07-15
net: remove all namehash usage

dr-frmr:
## Problem

We no longer need namehash indexing at the networking protocol level. This simplifies the data structures significantly and allows us to only care about actual node names at the runtime level.

## Solution

TODO 

## Testing

TODO

## Docs Update

TODO

## Notes

This is a **breaking change** affecting `net:distro:sys` API. Document accordingly on 0.9.0 release.


barraguda (approved):
very nice!

Issue #448, opened 2024-07-16
bug: can't import keyfile (0.8.3)


0x70b1a5:
**Describe the bug**
I cannot import a keyfile to a fresh node.

**To Reproduce**
1. Export your .keys from a node, `node1`.
2. Boot a fresh node at a new folder, `node2`.
3. Open browser and select your .keys file to import.
4. See error. Console reads 404 at /vet-keyfile.

**Expected behavior**
It should let me boot `node2` with the same identity as `node1`.

**Screenshots**
![image](https://github.com/user-attachments/assets/401b52b8-a158-45cf-b98c-089f5622b2bb)

**Desktop (please complete the following information):**
 - OS: MacBook Pro 2017, Ventura 13.6.7
 - Browser Brave
 - Kinode version 0.8.3
 - process_lib version unk

**Additional context**

dr-frmr:
fixed in #451

PR #451, opened 2024-07-18
v0.8.5

nick1udwig:


dr-frmr (changes requested):
Before merge, needs to have `0.8.5` as version across Cargo.tomls and should target `release-candidate` -- also I want to fix this #448 this morning too.

PR #455, opened 2024-07-18
sqlite: fix potential nullable columns read

barraguda:


nick1udwig:
Tested using https://github.com/kinode-dao/core_tests and worked as expected. Thanks for the fix :rocket:

nick1udwig (approved):

PR #457, opened 2024-07-19
eth: fix deadlock casued by multiple subscriptions/requests

nick1udwig:
## Problem

Our default eth provider, `providerfren.os`, frequently hangs, requiring restart.

## Solution

Fix deadlock by minimizing the amount of time locks are acquired for. In particular move everything that communicates with a timeout outside of a lock.

## Testing

To test I:
1. Set up an eth provider using this binary or master
2. Modified kns_indexer to infinitely loop sending `get_logs()`, by commenting out https://github.com/kinode-dao/kinode/blob/fbf961e9de224de922e274b08bfbf860311dd963/kinode/packages/kns_indexer/kns_indexer/src/lib.rs#L163
3. Run one node using looping kns indexer, pointed at eth provider
4. Start a second one & stop if it successfully connects, repeating until either witness deadlock or get bored

Using that protocol, I get a deadlock immediately on master, whereas I have not witnessed a deadlock on this branch.

## Docs Update

None

## Notes

None

### DIFF
--- a/kinode/src/eth/mod.rs
+++ b/kinode/src/eth/mod.rs
@@ -37,14 +37,14 @@ struct ActiveProviders {
     pub nodes: Vec<NodeProvider>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 struct UrlProvider {
     pub trusted: bool,
     pub url: String,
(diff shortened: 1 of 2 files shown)


dr-frmr (approved):
amazing

PR #468, opened 2024-08-05
kns: routers encoding update

barraguda:
## Problem

routers were temporarily inscribed on-chain as strings, separated by commas. hashes seem better and might be a fair  bit shorter for longer names.
## Solution

post raw nodehashes on chain instead, KNS indexes them, resolves the names and then gives them to net:distro:sys

## Docs

whitepaper update todo? [x]

### DIFF
--- a/kinode/packages/kns_indexer/kns_indexer/src/lib.rs
+++ b/kinode/packages/kns_indexer/kns_indexer/src/lib.rs
@@ -311,7 +311,7 @@ fn handle_log(our: &Address, state: &mut State, log: &eth::Log) -> anyhow::Resul
             };
 
             state.names.insert(child_hash.clone(), full_name.clone());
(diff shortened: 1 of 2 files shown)


dr-frmr:
@bitful-pannul there also appears to be a mismatch between how IP addresses are inscribed -- can you look into that?

dr-frmr (commented):
Everything here looks good. Merge conflict is probably because I deleted prints vs. comment out. They can remain commented I suppose. I'd like to investigate the IP decode on this PR too, then merge.

barraguda:
> @bitful-pannul there also appears to be a mismatch between how IP addresses are inscribed -- can you look into that?

no mismatch, just migrated nodes with old encoding.
but ~ip values could be done in 4/16 byte arrays instead of always 16 maybe: https://github.com/kinode-dao/kinode/pull/469

PR #474, opened 2024-08-08
eth provider fixes

dr-frmr:
## Problem

Eth provider works well for client, but provider still gets bloat of many stale subscriptions

## Solution

On client side, manually unsubscribe any time a subscription is dropped.

## Testing

Run a provider and a client for a few days


dr-frmr:
@nick1udwig can you quickly review this, merge to 0.8.7 if you see fit, and then we can go right to release-candidate from there?

nick1udwig (approved):
LGTM; compiles without issue; haven't tested

PR #486, opened 2024-08-14
build script output reruns skip UI

### LINKED ISSUE #480: bug: build.rs does not build new core packages unless you scorch earth
**Describe the bug**
When you add a new core package, you must clear all the previous build information and build from scratch: otherwise the package does not get built and the binary will fail to run because it will not be able to find the .wasm binary bundled in with it.

dr-frmr:
## Problem

Build.rs still wonky, see #480

## Solution

Attempt to both make build script maximally sensitive to updates while skipping rebuild if only build-outputs have changed since last build. Seems to fix #480 

## Testing

build runtime,
change elements,
use SKIP_BUILD_FRONTEND flag,
etc

## Docs Update

N/A

## Notes

Let's test this on all the platforms we develop on frequently

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ rustup target add wasm32-wasi
 rustup target add wasm32-wasi --toolchain nightly
 cargo install cargo-wasi
 
+# Install NPM so we can build frontends for "distro" packages.
+# https://docs.npmjs.com/downloading-and-installing-node-js-and-npm
+# If you want to skip this step, run cargo build with the environment variable SKIP_BUILD_FRONTEND=true
+
 # Build the runtime, along with a number of "distro" Wasm modules.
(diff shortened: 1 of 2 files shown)


nick1udwig (approved):
Doesn't seem to resolve #480 for me. I did:
```
kit f -r $PATH_TO_KINODE
cp -r ${PATH_TO_KINODE}/kinode/packages/docs ${PATH_TO_KINODE}/kinode/packages/mocks
# MacOs version of `sed`
sed -i '' 's/docs/mocks/' ${PATH_TO_KINODE}/kinode/packages/mocks/metadata.json

kit f -r $PATH_TO_KINODE
```
noting that it does not rebuild in the second `kit f`

Still -- it does seem like an improvement over current build.rs, hence approval :+1:

PR #492, opened 2024-08-19
pending notes

nick1udwig:
## Problem

When a Mint and a Note happen for the same node in the same block, they can arrive in two orders:
1. Mint -> note
2. Note -> mint

Currently, 1 works properly, but 2 doesn't.

## Solution

Store notes that don't have their mint processed yet and process in a subsequent block.

## Testing

TODO

## Docs Update

None

## Notes

None

### DIFF
--- a/kinode/packages/kns_indexer/kns_indexer/src/lib.rs
+++ b/kinode/packages/kns_indexer/kns_indexer/src/lib.rs
@@ -20,9 +20,6 @@ wit_bindgen::generate!({
     additional_derives: [serde::Deserialize, serde::Serialize],
 });
 
-type PendingNotes = BTreeMap<u64, HashMap<kimap::contract::Note, u8>>;
-const MAX_PENDING_ATTEMPTS: u8 = 3;
-
 #[cfg(not(feature = "simulation-mode"))]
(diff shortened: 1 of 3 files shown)


kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -326,72 +453,27 @@ fn handle_log(our: &Address, state: &mut State, log: &eth::Log) -> anyhow::Resul
...
                 return Err(anyhow::anyhow!("skipping invalid note: {note}"));
             }
 
-            let Some(node_name) = get_parent_name(&state.names, &node_hash) else {
+            let Some(_node_name) = get_parent_name(&state.names, &node_hash) else {
+                pending_notes
+                    .entry(log.block_number.unwrap())
+                    .or_insert(vec![])
+                    .push(decoded);
                 return Err(anyhow::anyhow!("parent node for note not found"));
```
nick1udwig: We should:
1. Remove this error (maybe replace with a debug print?), or
2. Change the error text

barraguda: great point!

PR #505, opened 2024-08-26
fakenet fixes

barraguda:
## Problem
New kimap indexing changes not fully working on fakenodes.

## Solution
Do 1 transaction only in src/fakenet.rs for registration, optimistically handle notes assuming their mint event came first.

## Notes

Thought we might've had to do a `--block-time 3` or similar argument to anvil in kit chain, but does not look like it (if we handle notes optimistically). App store publishing works on fakenet

### DIFF
--- a/kinode/packages/kns_indexer/kns_indexer/src/lib.rs
+++ b/kinode/packages/kns_indexer/kns_indexer/src/lib.rs
@@ -433,15 +433,22 @@ fn handle_log(
             if !kimap::valid_note(&note) {
                 return Err(anyhow::anyhow!("skipping invalid note: {note}"));
             }
-            if let Some(block_number) = log.block_number {
-                print_to_terminal(
-                    1,
-                    &format!("adding note to pending_notes for block {block_number}"),
-                );
-                pending_notes
-                    .entry(block_number)
-                    .or_default()
-                    .push((decoded, 0));
+            // handle note: if it precedes parent mint event, add it to pending_notes
+            if let Err(e) = handle_note(state, &decoded) {
+                if let Some(KnsError::NoParentError) = e.downcast_ref::<KnsError>() {
+                    if let Some(KnsError::NoParentError) = e.downcast_ref::<KnsError>() {
+                        if let Some(block_number) = log.block_number {
+                            print_to_terminal(
+                                1,
+                                &format!("adding note to pending_notes for block {block_number}"),
+                            );
+                            pending_notes
+                                .entry(block_number)
+                                .or_default()
+                                .push((decoded, 0));
+                        }
+                    }
+                }
             }
         }
         _log => {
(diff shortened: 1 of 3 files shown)


kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -433,15 +433,22 @@ fn handle_log(
...
-                    &format!("adding note to pending_notes for block {block_number}"),
-                );
-                pending_notes
-                    .entry(block_number)
-                    .or_default()
-                    .push((decoded, 0));
+            // handle note: if it precedes parent mint event, add it to pending_notes
+            if let Err(e) = handle_note(state, &decoded) {
+                if let Some(KnsError::NoParentError) = e.downcast_ref::<KnsError>() {
+                    if let Some(KnsError::NoParentError) = e.downcast_ref::<KnsError>() {
```
nick1udwig: Looks like an accidental duplicate here

nick1udwig:
image
can't `hi` other nodes

What I did:

```
# Checkout Kinode branch
git checkout bp/fakenet-fix

# Start one fake node
kit f -r ~/git/kinode

# Start another fake node
kit f -r ~/git/kinode -o /tmp/kinode-fake-node-2 -p 8081 -f fake2.dev

# Try to talk
hi fake.dev hello
```

nick1udwig:
Ok, this is getting more stable on my machine.

Next step: getting it to work with `run-tests`.

Currently does not work, I get:
```
$ kit t foo

...

stderr:
thread 'main' panicked at kinode/src/main.rs:574:18:
called `Result::unwrap()` on an `Err` value: server returned an error response: error code -32003: replacement transaction underpriced

...
```

To run this test do:

```
# Create template
kit n foo

# Edit the tests.toml config to point to your local repo (with this branch checked out), not to fetch the 0.9.1 binary
## Comment out the first line; uncomment the second & edit the path to point to your Kinode core repo
vi foo/test/tests.toml

# Run the test
kit t foo
```

barraguda:
> Ok, this is getting more stable on my machine.
> 
> Next step: getting it to work with `run-tests`.

@nick1udwig 
It should robustly work now, run_tests too, update to a new kit on branch `bp/chain-blocktime` and pull this branch too.

I have some notes on the kit PR that might be worth considering.

nick1udwig (approved):
It works! Thanks for the fixes.

Before merging, please quantify memory usage. If it doesn't look like it runs away after being run for an hour, good to merge; else, let's discuss more before merging.

barraguda:
> It works! Thanks for the fixes.
> 
> Before merging, please quantify memory usage. If it doesn't look like it runs away after being run for an hour, good to merge; else, let's discuss more before merging.

anvil, starting at 10mb mem-usage for me, does seem to runaway with ~5mb per min or so.

barraguda:
ok, small ~1.5h anvil memory monitoring plot here (just kit f running in background), goes up to over 100mb, then comes down and stabilizes. Anvil does some disk caching to /tmp, I assume it's that. 

There's another (hackier?) workaround that we could do for fakenodes. This is upon registering in fakenet.rs, we post 1wei of eth to the zero address. All this does is mine 1 extra block after registering.

I'm leaning towards that, having blocks constantly mined in the background seems slightly sketchy. 
image

PR #511, opened 2024-08-27
feat: let user reset any node name, not just `.os`

dr-frmr:
## Problem

Current login reset-keys flow is fixed to `.os` TLZ

## Solution

Add optional TLZ argument to name entry so that we can still fix TLZ for main registration flow, but user can reset any node ID they own.

## Testing

```
1. mint regular .os name to confirm fixed flow works (don't need to actually execute txn)
2. reset regular .os name (reset text box will contain full name including .os)
3. mint non-.os name (i used .kino)
4. reset non-.os name 
```

## Docs Update

N/A

## Notes

I considered fixing `.os` in the blue box next to name entry if user is resetting a .os name, but having the text box be consistently the full name on the reset page seemed easier to understand.


nick1udwig (approved):
LGTM :+1:

PR #512, opened 2024-08-27
app_store: new downloads page, sanity checks while publishing

barraguda:
## ux

Rendering all versions is confusing, and they got missorted every now and then. Instead, default to newest version, and have the ability to choose with a dropdown. 

Also, if devs are about to publish a package, without having some hash of their code-versions on their local node, they should be warned. Same with publishing an invalid package_name


barraguda:
closed by newer clean commit log

PR #517, opened 2024-08-29
app_store: use fixed vfs::open_dir

nick1udwig:
## Problem

A broken process_lib function led to some broken code here.

## Solution

Use [fixed process_lib code](https://github.com/kinode-dao/process_lib/pull/91/commits/631d346ccf879b854ebc9b47f4c2cd71cf2f1771) & fix code here.

## Testing

Does it work? Note master should work as well

## Docs Update

None

## Notes

None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,7 +78,7 @@ name = "alias"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "kinode_process_lib 0.9.0 (git+https://github.com/kinode-dao/process_lib?tag=v0.9.0)",
+ "kinode_process_lib 0.9.0",
  "serde",
  "serde_json",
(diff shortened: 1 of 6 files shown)


dr-frmr:
Can we put out process_lib 0.9.1 and upgrade *all* usages in runtime to that new version?

barraguda (approved):
nice thank you! tested app_store downloads interactions with this, lgtm

barraguda:
~~actually @nick1udwig , should we add open_or_create_dir() as a helper function to process_lib? seems like something we would use reasonably often.
Same with open_or_create_file()?~~

potential PR here: https://github.com/kinode-dao/process_lib/pull/92
can update usage in core too

edit: no new helpers needed, just a further adjustment in open_dir!

PR #519, opened 2024-08-30
eth: fix remote provider

nick1udwig:
## Problem

providerfren.os still freezes

## Solution

To solve the cause of the freeze:

1. dont unsubscribe, since subs are de-duped by alloy; if you unsub one mint, e.g., you unsub all
2. cache eth requests
3. retry eth requests once on rpc error

It is still unsettling that not handling unsubs properly leads to a freeze, and that needs to be run down, but these changes seem to solve the cause.

## Testing

See if providerfren.os is still alive (it is running this runtime version)

## Docs Update

None

## Notes

None

### DIFF
--- a/kinode/src/eth/mod.rs
+++ b/kinode/src/eth/mod.rs
@@ -727,7 +727,7 @@ async fn fulfill_request(
         let mut request_cache = request_cache.lock().await;
         if let Some((cache_hit, time_of_hit)) = request_cache.shift_remove(&serialized_action) {
             // refresh cache entry (it is most recently accessed) & return it
-            if time_of_hit.elapsed() > Duration::from_millis(DELAY_MS) {
+            if time_of_hit.elapsed() < Duration::from_millis(DELAY_MS) {
                 println!("cache hit\r");
                 request_cache.insert(serialized_action, (cache_hit.clone(), time_of_hit));
                 return cache_hit;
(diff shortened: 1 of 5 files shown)


kinode/src/eth/mod.rs
```diff
@@ -650,19 +714,36 @@ async fn fulfill_request(
...
     let EthAction::Request {
-        chain_id,
+        ref chain_id,
         ref method,
         ref params,
     } = eth_action
     else {
         return EthResponse::Err(EthError::PermissionDenied); // will never hit
     };
+    {
```
barraguda, line 726: I think there's a flaw with this logic:

Our current filters in kns & app_store aren't unique, they look identical ("get me all previous logs until *now*"). If one node serializes this, gets back a correct result -> it's cached, another node boots, we index it correctly but a third node asks for getLogs, misses the mint and gets the cache hit instead.

Here's what I observed (all nodes using providerfren.os): 
- boot monkmania.os (old node, just get up to date)
- mint biogastrus.os, shut down
- mint biomaximus.os
- boot biogastrus.os again, cannot talk to biomaximus.os
- monkmania.os has the peer of both new nodes (come in via subscriptions and not getLogs)

nick1udwig: Good catch. I believe I've fixed in 953f1ff. Solution is to just invalidate all cache entries after 1s.

barraguda (approved):
nice!

I think something we can do in the near future to make the caching more robust (and for potential further deduplication of eth requests) is to specify our get_logs requests more thoroughly.

Saving state in kns & app_store probably too, but mainly getting a block number at boot, then specifying from_block -> to_block in the filter.

PR #522, opened 2024-09-03
app_store: always respond with error; use process_macros

nick1udwig:
## Problem

1. app_store can error without sending a response, which can hang, e.g., `kit`
2. app_store doesn't make use of process_macros

## Solution

1. Add HandlingError & SendError variants
2. Use process_macros

## Testing

```
# Get `kit` https://github.com/kinode-dao/kit/pull/221/commits/f9f9ce8b056d9412e91f9401d2d32c56eb8699e2 or earlier
cargo install --git https://github.com/kinode-dao/kit --rev f9f9ce8

# Start a package & then remove it
kit n chat
kit b chat
kit s chat
kit r chat
```

`kit r` hangs because[ f9f9ce8](https://github.com/kinode-dao/kit/pull/221/commits/f9f9ce8b056d9412e91f9401d2d32c56eb8699e2) and earlier had a bug and app_store does not issue a Response and so HTTP server never issues an HTTP Response and so `kit` hangs.

## Docs Update

None

## Notes

process_macros required some reworking of variant names. Rust compiler got mad about `::Error` variants alongside [the `TryFrom` `type Error` definitions in process_macros (whcih are required for TryFrom)](https://github.com/kinode-dao/process_macros/blob/master/src/lib.rs#L24). Solved by changing `::Error` variant names to `::Err`. We should probably document this somewhere.

### DIFF
--- a/kinode/packages/app_store/app_store/src/lib.rs
+++ b/kinode/packages/app_store/app_store/src/lib.rs
@@ -74,25 +74,11 @@ fn init(our: Address) {
     loop {
         match await_message() {
             Err(send_error) => {
-                // TODO: handle these more gracefully so only real send errors are printed
-                // // for now, these are timer callbacks to already finished ft_workers.
-                // print_to_terminal(1, &format!("got network error: {send_error}"));
-
-                let error_message = format!("got network error: {send_error}");
-                println!("{error_message}");
-                Response::new()
-                    .body(AppStoreResponse::SendError(error_message))
-                    .send()
-                    .unwrap();
+                print_to_terminal(1, &format!("main: got network error: {send_error}"));
             }
             Ok(message) => {
                 if let Err(e) = handle_message(&our, &mut state, &mut http_server, &message) {
-                    let error_message = format!("error handling message: {e:?}");
-                    println!("{error_message}");
-                    Response::new()
-                        .body(AppStoreResponse::HandlingError(error_message))
-                        .send()
-                        .unwrap();
+                    print_to_terminal(1, &format!("main: error handling message: {e:?}"));
                 }
             }
         }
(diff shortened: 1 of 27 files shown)


nick1udwig:
@bitful-pannul ptal and in particular please look at `SendError`: I'm not sure what your comment there meant, and so if it meant we were using `SendError` as a catch for some type of common thing, we should figure out how to handle that frequent case.

I'm also not certain that the `SendError` Response will work properly anyways.

kinode/packages/app_store/app_store/src/lib.rs
```diff
@@ -73,12 +74,25 @@ fn init(our: Address) {
...
+                Response::new()
+                    .body(AppStoreResponse::SendError(error_message))
+                    .send()
+                    .unwrap();
             }
             Ok(message) => {
                 if let Err(e) = handle_message(&our, &mut state, &mut http_server, &message) {
-                    println!("error handling message: {:?}", e);
+                    let error_message = format!("error handling message: {e:?}");
+                    println!("{error_message}");
```
barraguda: are we sure we want to print with full vebosity here? I'll test with some basic interactions to see if it doesn't muddle up terminal too much

nick1udwig: Not sure about it. Previous behavior was verbosity 0, so just maintaining that. Happy to change.

barraguda: I set it it to verbosity 1 and removed the response, it's not really ever hit or listened to. Added some clarifications to the verbosity prints of chain and downloads so one knows where they come from.

barraguda (approved):
process_macros are sick!

nick1udwig:
Ty @bitful-pannul :rocket:

PR #528, opened 2024-09-06
terminal: refactor to use unicode graphemes properly, fix special character bugs

dr-frmr:
## Problem

The terminal was using string slicing to do all character-level manipulation, causing special characters to create panics due to inserts/removes in the middle of unicode graphemes.

## Solution

Rewrite the terminal with the `unicode_segmentation` crate to properly handle character input.

## Testing

```
1. Use terminal as usual (major refactor, so test old behaviors)
2. Use terminal with special characters!
```

## Docs Update

N/A

## Notes

N/A

PR #529, opened 2024-09-06
terminal: dont crash if no size

nick1udwig:
## Problem

Terminal crashes Kinode core when running in CI workflow

image

## Solution

Don't crash.

## Testing

It builds; haven't tested in CI yet.

## Docs Update

None

## Notes

None


dr-frmr:
Might as well do this in terminal/mod.rs line 174 too. Also I did a big refactor of terminal here: #528

dr-frmr (approved):

PR #530, opened 2024-09-06
make docs app serve http dynamically

dr-frmr:
## Problem

Docs app was making a ton of static bindings, leading to startup-cruft and excess memory usage (not truly an issue, but this way is *cleaner*)

## Solution

Dynamically serve docs as HTTP requests occur

## Testing

Use the app!

## Docs Update
N/A
## Notes
N/A

### DIFF
--- a/kinode/packages/docs/docs/src/lib.rs
+++ b/kinode/packages/docs/docs/src/lib.rs
@@ -1,8 +1,5 @@
 use kinode_process_lib::{
-    await_message, call_init,
-    homepage::add_to_homepage,
-    http::server::{HttpBindingConfig, HttpServer},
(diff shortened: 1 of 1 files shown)


nick1udwig (approved):
LGTM, thanks for the change.

One thing this makes me wonder about is whether we can make use of lazy load/inheritance with these different fancy methods for FS, etc. Its tricky because we can't assume what user is going to do after the method call.

Maybe it is fine: as of now, default is not 100% performant but is ergonomic; if you want 100% performance, you have to write it yourself using inheritance.

OTOH it would be v nice if we could somehow "do it for" people. I expect it'd be tricky tho.

Issue #525, opened 2024-09-04
bug: still getting "Too many open files" OS error


nick1udwig:
**Describe the bug**
Can sometimes get an OS error from linux: "Too many open files"

https://discord.com/channels/1186394868336038080/1186684706700410962/1280889566110289922

**To Reproduce**
Not sure

**Expected behavior**
Don't open too many files

nick1udwig:
https://discord.com/channels/1186394868336038080/1186684706700410962/1281722686787289178

barraguda:
to reproduce one can write a once running process something like this

I hit the limit at ~200 files. Working on a forced cleanup if MAX_FILES are reached rather than just running one at an interval.

```rust
fn init(our: Address) {
    println!("begin");

    let drive = vfs::create_drive(our.package_id(), "hello", None).unwrap();
    let mut open_files: HashMap<String, File> = HashMap::new();

    let amount_files = 500;
    for i in 0..amount_files {
        let file_name = format!("{}/test_{}.txt", &drive, i);
        match vfs::open_file(&file_name, true, None) {
            Ok(mut file) => {
                match file.write_all("hellO world".as_bytes()) {
                    Ok(_) => {}
                    Err(e) => println!("got error while writing to file {}: {e:?}", file_name),
                }
                open_files.insert(file_name.clone(), file);
            }
            Err(e) => println!("got error while opening file: {e:?}"),
        }
    }
    println!("wrote {} files", amount_files);

    for (file_name, file) in open_files.iter_mut() {
        match file.read_to_string() {
            Ok(s) => println!("read: {s}, from {file_name}"),
            Err(e) => println!("got error while reading file {}: {e:?}", file_name),
        }
        match file.write_all("hellO world".as_bytes()) {
            Ok(_) => {}
            Err(e) => println!("got error while writing to file {}: {e:?}", file_name),
        }
    }

    println!("done.");
}
```

PR #533, opened 2024-09-10
vfs: cleanup on max_open_files threshold

barraguda:
## Problem

https://github.com/kinode-dao/kinode/issues/525

## Solution

If open_files.len() is above threshold (180 right now based on my mac testing), run a cleanup that prunes 50% of the open files. 

Interval cleanup of idle files still happens, but now in a tokio::select loop instead.

## Notes

Linux seems to have a larger supported open filehandles amount per OS process (~1000 as opposed to mac's 256)

### DIFF
--- a/kinode/src/vfs.rs
+++ b/kinode/src/vfs.rs
@@ -1,4 +1,4 @@
-use dashmap::{mapref::entry::Entry, DashMap};
+use dashmap::DashMap;
 use lib::types::core::{
     Address, CapMessage, CapMessageSender, Capability, DirEntry, FileMetadata, FileType,
     KernelMessage, LazyLoadBlob, Message, MessageReceiver, MessageSender, PackageId, PrintSender,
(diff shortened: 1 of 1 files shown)


kinode/src/vfs.rs
```diff
@@ -594,26 +593,49 @@ async fn open_file<P: AsRef<Path>>(
...
+        .iter()
+        .map(|entry| (entry.key().clone(), entry.value().1))
+        .collect();
+    entries.sort_by(|a, b| a.1.cmp(&b.1));
+
+    // remove oldest entries until we reach the target count
+    for (path, _) in entries.into_iter().take(current_count - target_count) {
+        // use remove_if_present to avoid potential race conditions
+        open_files.remove_if(&path, |_, _| true);
+    }
```
dr-frmr: is there a way to make this more efficient using `retain` on the dashmap with a fixed duration calculated by the average of the map? could be better than cloning every path. something like

1. calculate average last access time by traversing map once
2. retain entries only with a below average last access time

barraguda: this is a good point, I'll try something out!

barraguda:
some notes for the other suggestions during discussions: 

1. matching upon a specific "TooManyOpenFiles" error in the open_file() function seems to me slightly too risky, there's no specific error kind like that in tokio::fs and matching on raw OS error codes that may be differ in environments seems risky
2. closing 1 random file every time we reach the limit rather than 50% of them has around the same performance for 5000 sequential file opens (~700ms-900ms on my machine, debug binary)

nick1udwig:
1. Surprised about this. I agree matching on raw OS error codes seems less good
2. Gotta test performance on release; debug is meaningless

dr-frmr:
Possible/reasonable to match on *any* error coming out of open_file()?

barraguda:
> Possible/reasonable to match on _any_ error coming out of open_file()?

I like this idea, but noting that keeping the open_files very close to the limit and relying on specific errors like this might be brittle. 

Another big note is, we need to save cursor positions of files we close, so that processes that expect them to be at a certain location won't notice a difference even if the vfs closed and reopened them. This is ok but leads to us not being able to use synchronous functions like dashmap.retain() and makes closing a lot of files considerable slower..

dr-frmr (approved):
Nice refactor, looks clean

PR #540, opened 2024-09-16
homepage: visuals

barraguda:
@wistec-bisrum


kinode/packages/homepage/ui/src/index.css
```diff
@@ -1,226 +1,392 @@
+@import url("https://api.fontshare.com/v2/css?f[]=clash-display@400,700,500,600,300&display=swap");
```
dr-frmr, line 1: Is this font open-source etc?

kinode/packages/homepage/ui/src/index.css
```diff
@@ -1,226 +1,392 @@
...
+}
+
+@media (max-width: 768px) {
+  body,
+  html {
+    height: 100%;
+    margin: 0;
+    padding: 0;
+    overflow: hidden;
+  }
```
dr-frmr, line 214: How much of this (and other stuff in this file) makes sense to expose globally in kinode.css?

dr-frmr (approved):
Visually I really like this. Would be happy to merge and work out what belongs in kinode.css later. @bitful-pannul what do you think?

EDIT: let's change font back to Kode Mono and remove the import added here. Team likes old font more

kinode/packages/homepage/ui/src/index.css
barraguda: imo assuming a body structure and things like mobile width adjustments might be too opinionated if people want to pull in kinode.css into their UIs

PR #542, opened 2024-09-16
register flow: add back button to UI where relevant

dr-frmr:
## Problem

Users in the desktop app don't have regular browser controls to navigate backwards through register flow.

## Solution

Add an in-UI back button to register/login pages where potentially useful.

## Testing

1. navigate naturally through register flow, trying various options

## Docs Update

N/A

## Notes

Visually:
Screenshot 2024-09-16 at 3 11 55 PM

### DIFF
--- a/kinode/src/register-ui/src/index.css
+++ b/kinode/src/register-ui/src/index.css
@@ -244,3 +244,10 @@
 button.secondary {
     width: 100%;
 }
+
+button.back {
+    float: left;
+    width: auto !important;
+    padding: 5px !important;
+    border: none !important;
+}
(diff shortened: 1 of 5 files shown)


barraguda (commented):
Nice, needed this. I wonder if it would be even clearer if there were a (nonclickable) progress bar with dots to indicate which part of the registration process a given user is on:

`[x] commit to a name ---> [x] register name ---> [] set password`

dr-frmr:
> Nice, needed this. I wonder if it would be even clearer if there were a (nonclickable) progress bar with dots to indicate which part of the registration process a given user is on:
> 
> `[x] commit to a name ---> [x] register name ---> [] set password`

Creating this as an issue

PR #543, opened 2024-09-17
app_store: docstrings

barraguda:
## Problem

Many separate processes with different responsibilities in app_store.

## Solution

document them well!

## Notes

slight WIP: might be more verbose in places where it's not needed, and not verbose enough in others.


dr-frmr (approved):
Very nice.

PR #546, opened 2024-09-18
add logging

nick1udwig:
## Problem

It would be nice to have logging for processes.

## Solution

Implement some helper functions for `tracing` to make it easy to log.

## Testing

TODO

## Docs Update

TODO

## Notes

Connected to https://github.com/kinode-dao/process_lib/pull/96

Note that this PR is not strictly necessary. The entirety of changes could be in userspace if instead of using WASI to write to the `.log` file we used the vfs:

https://github.com/kinode-dao/process_lib/blob/acc4a67b594d3042c4cc66157b6cce397d73ac77/src/logging.rs#L28-L40

Considering making the change.


nick1udwig:
Decided to use VFS instead of WASI

PR #549, opened 2024-09-18
kernel: output import/export errors

### LINKED ISSUE #548: bug: deadlock/crash process when `send_and_await()` to `our`
**Describe the bug**
Can deadlock/crash process when `send_and_await()`ing to `our`.

**To Reproduce**
`send_and_await()` to `our`.

**Expected behavior**
Either
1. Refuse to `send_and_await()` to `our` & print a useful error message for devs, or
2. Handle this case in a clever way somehow.

**Additional context**
h/t @0x70b1a5 
https://discord.com/channels/1186394868336038080/1186421052071477428/1286011373960040518

nick1udwig:
## Problem

1. Resolves #548
2. Kernel import/export errors not output

## Solution

1. Return helpful error if `target == our` for `send_and_await()`
2. If no `stderr` from process, print `call_init()`s error root cause

## Testing

Run a process that `send_and_await_response()`s to itself on master and here.

## Docs Update

None

## Notes

Error print looks like:
```
Wed 14:13 process diligence.os@hq:hq:uncentered.os ended with error:
kernel: got invalid send_and_await_response() Request from and to ProcessId { process_name: "hq", package_name: "hq", publisher_node: "uncentered.os" }: cannot await a Request to `our`: will deadlock
```

### DIFF
--- a/kinode/src/kernel/process.rs
+++ b/kinode/src/kernel/process.rs
@@ -285,12 +285,17 @@ pub async fn make_process_loop(
                         .send(&send_to_terminal)
                         .await;
                 }
-                Err(_) => {
+                Err(e) => {
                     let stderr = wasi_stderr.contents().into();
                     let stderr = String::from_utf8(stderr)?;
+                    let output = if stderr != String::new() {
+                        stderr
+                    } else {
+                        format!("{}", e.root_cause())
+                    };
                     t::Printout::new(
                         0,
-                        format!("\x1b[38;5;196mprocess {our} ended with error:\x1b[0m\n{stderr}",),
+                        format!("\x1b[38;5;196mprocess {our} ended with error:\x1b[0m\n{output}"),
                     )
                     .send(&send_to_terminal)
                     .await;
@@ -313,12 +318,17 @@ pub async fn make_process_loop(
                         .send(&send_to_terminal)
                         .await;
                 }
-                Err(_) => {
+                Err(e) => {
                     let stderr = wasi_stderr.contents().into();
                     let stderr = String::from_utf8(stderr)?;
+                    let output = if stderr != String::new() {
+                        stderr
+                    } else {
+                        format!("{}", e.root_cause())
+                    };
                     t::Printout::new(
                         0,
-                        format!("\x1b[38;5;196mprocess {our} ended with error:\x1b[0m\n{stderr}",),
+                        format!("\x1b[38;5;196mprocess {our} ended with error:\x1b[0m\n{output}"),
                     )
                     .send(&send_to_terminal)
                     .await;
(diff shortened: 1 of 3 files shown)


kinode/src/kernel/process.rs
```diff
@@ -285,12 +285,17 @@ pub async fn make_process_loop(
...
                 }
-                Err(_) => {
+                Err(e) => {
                     let stderr = wasi_stderr.contents().into();
                     let stderr = String::from_utf8(stderr)?;
+                    let output = if stderr != String::new() {
+                        stderr
+                    } else {
+                        format!("{}", e.root_cause())
+                    };
```
dr-frmr: Very nice.

kinode/src/kernel/standard_host.rs
```diff
@@ -402,6 +402,12 @@ async fn send_and_await_response(
             process.process.metadata.our.process
         ));
     }
+    if t::Address::de_wit(target.clone()) == process.process.metadata.our {
+        return Err(anyhow::anyhow!(
+            "kernel: got invalid send_and_await_response() Request from and to {:?}: cannot await a Request to `our`: will deadlock",
```
dr-frmr: Personally I would prefer to get the pretty Display print of the `Address` rather than the Debug print

kinode/src/kernel/standard_host_v0.rs
```diff
@@ -409,6 +409,12 @@ async fn send_and_await_response(
             process.process.metadata.our.process
         ));
     }
+    if t::Address::de_wit_v0(target.clone()) == process.process.metadata.our {
+        return Err(anyhow::anyhow!(
+            "kernel: got invalid send_and_await_response() Request from and to {:?}: cannot await a Request to `our`: will deadlock",
```
dr-frmr: Same here regarding Display vs. Debug

dr-frmr (commented):
LGTM; I would suggest going to Display for `Address` if it's all the same to you.

kinode/src/kernel/standard_host.rs
nick1udwig: Good call; changed in 06c0bd4

kinode/src/kernel/standard_host_v0.rs
nick1udwig: 06c0bd4

kinode/src/kernel/process.rs
nick1udwig: Made me realize we can use `.is_empty()` here in the conditional logic; 06c0bd4

nick1udwig:
Ah, just going to merge since I addressed your comments & saw the "LGTM"

PR #550, opened 2024-09-19
app store: only do work to populate apps once

dr-frmr:
## Problem

Currently the app store tries to fetch a bunch of metadata during initial log ingestion, leading to unnecessary errors

## Solution

Only fetch metadata if we are ingesting a log *after* the startup fetch. After the startup fetch, do another pass to fetch metadata for each listing once

## Testing

```
1. start node
2. check app store
```

## Docs Update

N/A

## Notes

I think we'll want to go back to storing some data on disk soon so we don't need to re-fetch logs on every boot. We've made things more robust since we removed caching.

### DIFF
--- a/kinode/packages/app_store/app_store/src/http_api.rs
+++ b/kinode/packages/app_store/app_store/src/http_api.rs
@@ -45,12 +45,7 @@ pub fn init_frontend(our: &Address, http_server: &mut server::HttpServer) {
             .expect("failed to bind http path");
     }
     http_server
-        .serve_ui(
-            &our,
-            "ui",
-            vec!["/", "/app/:id", "/publish", "/download/:id", "my-downloads"],
-            config.clone(),
-        )
+        .serve_ui(&our, "ui", vec!["/"], config.clone())
         .expect("failed to serve static UI");
 
     http_server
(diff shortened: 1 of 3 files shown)


nick1udwig (approved):
LGTM; app_store looks to be properly set up after boot :+1:

kinode/packages/app_store/app_store/src/http_api.rs
```diff
@@ -45,12 +45,7 @@ pub fn init_frontend(our: &Address, http_server: &mut server::HttpServer) {
             .expect("failed to bind http path");
     }
     http_server
-        .serve_ui(
```
barraguda, line 48: nice! I thought this was necessary for users being able to go to a direct `/app/yourapp` link, but it works with just the "/" base path too.

barraguda (approved):
great stuff, good way to avoid time and repeated http calls to outdated apps. Let's enable state saving soon too

kinode/packages/app_store/app_store/src/http_api.rs
dr-frmr: it used to be, but now http server will just serve what's on the base path to unbound paths!

PR #551, opened 2024-09-19
app_store UI: don't persist local UI store

barraguda:
## Problem

After resetting a node, I somehow saw an "installed" button when trying to download an app, this should never happen. 

Our apis already fetch things upon load, so persisting them in localStorag for "quicker" caches doesn't really make sense.

Also, after installing an app, we do a fetch to homepages `/apps` endpoint to fetch the newly installed apps potential UI path. There's a race condition present here where the app is installed, but hasn't yet added itself to the homepage! 

## Solution
Do not persist the zustand store. 

For the launchbutton, I've added a slightly hacky 500ms sleep after installing and before fetching new homepageApps. Could be done by a timeout/backoff strategy but this is a good start.

-- edit:

Added a defaulting action for mirrors, if publisher not offline, choose the next online mirror if available.

## Testing

```
# Follow these steps to test
blank node, go through download/install flow for an app with UI, make sure the launch button properly shows up after installing.
```

### DIFF
--- a/kinode/packages/app_store/ui/src/store/index.ts
+++ b/kinode/packages/app_store/ui/src/store/index.ts
@@ -45,361 +45,361 @@ interface AppsStore {
 
 }
 
-const useAppsStore = create<AppsStore>()(
-  persist(
-    (set, get): AppsStore => ({
-      listings: {},
-      installed: {},
-      downloads: {},
-      ourApps: [],
-      activeDownloads: {},
-      homepageApps: [],
+const useAppsStore = create<AppsStore>()((set, get) => ({
+  listings: {},
+  installed: {},
+  downloads: {},
+  ourApps: [],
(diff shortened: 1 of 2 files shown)


kinode/packages/app_store/ui/src/store/index.ts
```diff
@@ -45,361 +45,361 @@ interface AppsStore {
...
+        // to give the app time to add itself to the homepage
+        // might make sense to add more state and do retry logic instead.
+        await new Promise(resolve => setTimeout(resolve, 500));
+
+        await get().fetchHomepageApps();
+      }
+    } catch (error) {
+      console.error("Error installing app:", error);
+    }
+  },
```
barraguda, line 232: sleep added here

dr-frmr (approved):

PR #558, opened 2024-09-24
terminal: add log rotation by default

nick1udwig:
## Problem

Terminal log balloons infinitely

## Solution

Add log rotation by default. Default behavior is to have a maximum total log size of 1GB spread across 4 logs. When a new log is created, remove oldest log.

## Testing

```
# Create package that will loop prints to logs

kit n foo

# Edit to add
# `
# loop {
#     println!("lol");
# }
# `
vi foo/foo/src/lib.rs

# Run it
kit f -r ~/git/kinode

# Enable logging: Ctrl + L

# Install
kit bs foo
```

## Docs Update

TODO

## Notes

Context: https://discord.com/channels/1186394868336038080/1186421052071477428/1288173581083082793 and above

Should logging be on by default? I tend to think yes

Current settings are for 1GB total logs on disk split b/w four files

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ The `sys` publisher is not a real node ID, but it's also not a special case valu
 - CTRL+J to toggle debug mode
 - CTRL+S to step through events in debug mode
 
-- CTRL+L to toggle logging mode, which writes all terminal output to the `.terminal_log` file. Off by default, this will write all events and verbose prints with timestamps.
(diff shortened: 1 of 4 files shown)


nick1udwig:
Note there is currently no way to change the setting from rotating to infinite. Ctrl+L still has the same behavior as before: turn on vs turn off file logging.

dr-frmr (approved):
Looks perfect. I notice that log will catch *all* verbosities regardless of terminal verbosity setting, but what happens to event-loop level print? No log of that, I guess?

Issue #556, opened 2024-09-24
feature: improve "file" handle managment by incorporating non-file "files" like sockets


nick1udwig:
**Is your feature request related to a problem? Please describe.**
We run out of unix "file" handles.

**Describe the solution you'd like**
Don't run out of "file" handles.

We currently track & clean up open files in VFS, but don't do the same for other modules that consume "file" handles, e.g. `net`.

PR #561, opened 2024-09-26
add `fd_manager` runtime module

nick1udwig:
## Problem

We run out of file descriptors because things like TCP sockets consume them: #556

## Solution

Introduce a new runtime module: `fd_manager`. It is pretty simple: it just tracks the number allowed `fd`s. Runtime modules that consume `fd`s must notify `fd_manager` when they open or close an `fd`. Then, if the total amount of `fd`s passes some threshold, `fd_manager` sends out a Request to all runtime modules with open `fd`s to `Cull` their old `fd`s with a suggested fraction to cull. Runtime modules that consume `fd`s must, therefore, also implement logic to respond to a `Cull`.

## Testing

TODO

## Docs Update

TODO

## Notes

TODO list:

- [x] Write `fd_manager` skeleton
- [x] Write `fd_manager` helper `send_fd_manager_open(number_opened: u64)`
- [x] Write `fd_manager` helper `send_fd_manager_close(number_closed: u64)`
- [x] Update vfs to talk to `fd_manager`
- [x] Update sqlite to talk to `fd_manager`
- [x] Update kv to talk to `fd_manager`
- [x] Update net to talk to `fd_manager`
- [ ] Look for any other consumers of fds and update them to talk to `fd_manager`
- [ ] Test

### DIFF
--- a/kinode/src/fd_manager.rs
+++ b/kinode/src/fd_manager.rs
@@ -267,11 +267,7 @@ async fn update_max_fds(state: &mut State) -> anyhow::Result<()> {
     Ok(())
 }
 
-async fn send_cull(
-    our_node: &str,
-    send_to_loop: &MessageSender,
-    state: &State,
-) {
+async fn send_cull(our_node: &str, send_to_loop: &MessageSender, state: &State) {
     let message = Message::Request(Request {
         inherit: false,
         expects_response: None,
--- a/kinode/src/fd_manager.rs
+++ b/kinode/src/fd_manager.rs
@@ -163,12 +163,20 @@ async fn handle_message(
             None
         }
         FdManagerRequest::CloseFds { mut number_closed } => {
-            assert!(state.total_fds >= number_closed);
             let mut return_value = Some(format!(
                 "{} closed {} of {}",
                 km.source.process, number_closed, state.total_fds,
             ));
-            state.total_fds -= number_closed;
+            if state.total_fds < number_closed {
+                return_value.as_mut().unwrap().push_str(&format!(
+                    "\n!!process claims to have closed more fds ({}) than we have open for all processes ({})!!",
+                    number_closed,
+                    state.total_fds,
+                ));
+                state.total_fds = 0;
+            } else {
+                state.total_fds -= number_closed;
+            }
             state
                 .fds
                 .entry(km.source.process)
--- a/kinode/src/fd_manager.rs
+++ b/kinode/src/fd_manager.rs
@@ -3,19 +3,23 @@ use lib::types::core::{
     MessageReceiver, MessageSender, PrintSender, Printout, ProcessId, Request,
     FD_MANAGER_PROCESS_ID,
 };
+use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, sync::Arc};
 
 const DEFAULT_MAX_OPEN_FDS: u64 = 180;
-const DEFAULT_FDS_AS_FRACTION_OF_ULIMIT_PERCENTAGE: u64 = 60;
+const DEFAULT_FDS_AS_FRACTION_OF_ULIMIT_PERCENTAGE: u64 = 90;
+const SYS_RESERVED_FDS: u64 = 30;
 const DEFAULT_UPDATE_ULIMIT_SECS: u64 = 3600;
 const _DEFAULT_CULL_FRACTION_DENOMINATOR: u64 = 2;
 
+#[derive(Debug, Serialize, Deserialize)]
 struct State {
     fds_limits: HashMap<ProcessId, FdsLimit>,
     mode: Mode,
     max_fds: u64,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
 enum Mode {
     /// don't update the max_fds except by user input
     StaticMax,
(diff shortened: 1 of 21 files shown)


kinode/src/fd_manager.rs
```diff
@@ -0,0 +1,274 @@
...
+    fn new() -> Self {
+        Self::default()
+    }
+
+    fn default() -> Self {
+        Self {
+            fds: HashMap::new(),
+            mode: Mode::default(),
+            total_fds: 0,
+            max_fds: DEFAULT_MAX_OPEN_FDS,
```
dr-frmr: Shouldn't `update_max_fds` be called once here at the start, so default limit is not applied briefly at launch?

nick1udwig:
When booting a pre-existing indirect node with this binary I got a segfault. Not obvious to me where in the `net` code there is a hanging `peer`, but that's what this looks like to me:
```
Tue 09:51 net: connection lost with aficionad.os
Tue 09:51 net:distro:sys closed 1 of 20
diligence.os > m our@fd_manager:distro:sys "GetState"
Tue 09:52 process diligence.os@12665670855602396235:terminal:sys returned without error
Tue 09:52 process 12665670855602396235:terminal:sys has OnExit behavior None
Tue 09:52 kernel: killing process 12665670855602396235:terminal:sys
Tue 09:52 hq:uncentered.os: error: Internal sync failed: Failed to generate embeddings: Timeout
Tue 09:53 net: attempting to connect to aficionad.os directly
Tue 09:53 net: connected to aficionad.os directly
diligence.os > thread 'tokio-runtime-worker' panicked at kinode/src/net/connect.rs:11:30:
                                                                                         net: peer sender was dropped: SendError { .. }
              Cleaning up "/home/nick/kinode-homes/diligence.os"...
Done cleaning up "/home/nick/kinode-homes/diligence.os".
```
Will look more into this this afternoon

nick1udwig:
Pretty sure I got it; confirming then fix coming here.

nick1udwig:
Fixed in aca7985. Unrelated fixes in a2fa276

dr-frmr:
By my count, the kernel / runtime itself uses about 20 file descriptors in the base case. I suggest that instead of configuring kernel, sqlite, kv, and terminal to use fd_manager, we simply add a fixed buffer that reserves the runtime FDs. Only vfs and net dynamically use a bunch of FDs / have the potential to overwhelm a limit at the moment.

Currently fd_manger uses a constant to give cushion from what the system gives:
```
const DEFAULT_FDS_AS_FRACTION_OF_ULIMIT_PERCENTAGE: u64 = 60;
```

I suggest that this gets bumped up to 90 and we tack on `30` as a fixed value that gets subtracted from whatever system ulimit we are given. If we are given less than 40 from system (which would give `9` as the final amount available to vfs/net.. not much to work with..) we crash with a descriptive error.

barraguda:
> I suggest that instead of configuring kernel, sqlite, kv, and terminal to use fd_manager, we simply add a fixed buffer that reserves the runtime FDs. Only vfs and net dynamically use a bunch of FDs / have the potential to overwhelm a limit at the moment.

This doesn't necessarily have to be in 0.9.5 yet, but kv and sqlite can definitely use a bunch of FDs (loop open 500 sqlite dbs will error), if users have many apps installed that use them, should be quite simple to cull them given most open actions are already OpenOrCreate.

nick1udwig:
I'm ok with hardcoding stuff for now to get a functional version out for our users, but definitely would want to have a less-hardcoded solution in the future.

dr-frmr:
> kv and sqlite FDs can definitely use a bunch of FDs

Wait really? I thought that both of these used one "db" and parceled userspace requests into separate tables and such.. if a new file descriptor is acquired we should def track

dr-frmr:
I see now that multiple DBs get opened. I'll wire these up to fd_manager.

dr-frmr:
@bitful-pannul need to you check my work here on refactor to `kv`. `sqlite` coming next

dr-frmr:
Sqlite work added.

kinode/src/sqlite.rs
```diff
@@ -20,6 +22,82 @@ lazy_static::lazy_static! {
...
+        access_order.push_back((package_id, db));
+        Ok(())
+    }
+
+    pub async fn remove_db(&mut self, package_id: PackageId, db: String) {
+        let db_path = format!("{}/{}/{}", self.sqlite_path.as_str(), package_id, db);
+        self.open_dbs.remove(&(package_id.clone(), db.to_string()));
+        let mut access_order = self.access_order.lock().await;
+        access_order.remove(&(package_id, db));
+        let _ = fs::remove_dir_all(&db_path).await;
```
barraguda: this removes the actual db, should just close the in-memory connection

kinode/src/kv.rs
```diff
@@ -11,6 +13,80 @@ use std::{
...
+        access_order.push_back((package_id, db));
+        Ok(())
+    }
+
+    pub async fn remove_db(&mut self, package_id: PackageId, db: String) {
+        let db_path = format!("{}/{}/{}", self.kv_path.as_str(), package_id, db);
+        self.open_kvs.remove(&(package_id.clone(), db.to_string()));
+        let mut access_order = self.access_order.lock().await;
+        access_order.remove(&(package_id, db));
+        let _ = fs::remove_dir_all(&db_path).await;
```
barraguda: +1, removes the db instead of closing the connection

barraguda (changes requested):

kinode/src/sqlite.rs
```diff
@@ -20,6 +22,82 @@ lazy_static::lazy_static! {
...
+
+        let db_conn = Connection::open(db_file_path)?;
+        let _ = db_conn.execute("PRAGMA journal_mode=WAL", []);
+
+        self.open_dbs.insert(key, Mutex::new(db_conn));
+
+        let mut access_order = self.access_order.lock().await;
+        access_order.push_back((package_id, db));
+        Ok(())
+    }
```
barraguda: should also update access order even if open_dbs.contains_key() returns true?

kinode/src/sqlite.rs
```diff
@@ -513,6 +514,12 @@ async fn check_caps(
             state
                 .remove_db(request.package_id.clone(), request.db.clone())
                 .await;
+
+            let _ = fs::remove_dir_all(format!(
+                "{}/{}/{}",
+                state.sqlite_path, request.package_id, request.db
+            ));
```
nick1udwig: We should never swallow `Result`s like this. It makes debugging really hard if an error ever starts being emitted. Either `unwrap()` it or -- at a minimum -- log it

PR #565, opened 2024-10-04
MVP of limit-based fd_manager

dr-frmr:
## Problem

Current fd_manager PR has concurrency issues; described in discord starting here: https://discord.com/channels/1186394868336038080/1186684706700410962/1291818971653345300

## Solution

Refactor fd_manager to allow individual runtime modules to maintain their own limits

## Testing

TODO; use https://github.com/dr-frmr/viifs

### DIFF
--- a/kinode/src/net/mod.rs
+++ b/kinode/src/net/mod.rs
@@ -71,7 +71,7 @@ pub async fn networking(
         active_passthroughs,
         max_peers,
         max_passthroughs,
-        fds_limit: 100, // TODO blocking request to fd_manager to get max num of fds at boot
+        fds_limit: 10, // small hardcoded limit that gets replaced by fd_manager soon after boot
     };
 
     let mut tasks = JoinSet::<anyhow::Result<()>>::new();
@@ -349,9 +349,10 @@ async fn handle_fdman(km: &KernelMessage, request_body: &[u8], data: &mut NetDat
             if data.max_passthroughs > fds_limit {
                 data.max_passthroughs = fds_limit;
             }
-            // TODO cull passthroughs too?
+            // TODO cull passthroughs too
             if data.peers.peers().len() >= data.fds_limit as usize {
-                data.peers.cull(2).await;
+                let diff = data.peers.peers().len() - data.fds_limit as usize;
+                data.peers.cull(diff).await;
             }
         }
         _ => return,
--- a/kinode/src/net/mod.rs
+++ b/kinode/src/net/mod.rs
@@ -1,6 +1,9 @@
-use lib::types::core::{
-    Identity, KernelMessage, MessageReceiver, MessageSender, NetAction, NetResponse,
-    NetworkErrorSender, NodeRouting, PrintSender,
+use lib::{
+    core::Address,
+    types::core::{
+        Identity, KernelMessage, MessageReceiver, MessageSender, NetAction, NetResponse,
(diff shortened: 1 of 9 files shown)


kinode/src/net/mod.rs
```diff
@@ -62,6 +71,7 @@ pub async fn networking(
         active_passthroughs,
         max_peers,
         max_passthroughs,
+        fds_limit: 100, // TODO blocking request to fd_manager to get max num of fds at boot
```
nick1udwig: seems too high -- maybe something more like 10 or 20?

kinode/src/net/mod.rs
```diff
@@ -322,20 +332,27 @@ async fn handle_local_request(
...
+                data.max_peers = fds_limit;
+            }
+            // TODO combine with max_peers check
+            if data.max_passthroughs > fds_limit {
+                data.max_passthroughs = fds_limit;
+            }
             // TODO cull passthroughs too?
-            data.peers.cull(cull_fraction_denominator).await;
+            if data.peers.peers().len() >= data.fds_limit as usize {
+                data.peers.cull(2).await;
```
nick1udwig: this doesn't seem quite right: seems like cull needs to be based on the difference to guarantee we are within the limit?

kinode/src/net/types.rs
```diff
@@ -103,13 +103,15 @@ impl Peers {
                 .key()
                 .clone();
             self.peers.remove(&oldest);
```
nick1udwig, line 105: doesn't need to be in this PR, but we should write an issue or a TODO to ourselves to make net & vfs cull in a more similar way (unless there is a specific reason to have them work differently)

kinode/src/vfs.rs
```diff
@@ -170,6 +165,7 @@ impl Files {
             access_order: Arc::new(Mutex::new(UniqueQueue::new())),
             our,
             send_to_loop,
+            fds_limit: 100, // TODO blocking request to fd_manager to get max num of fds at boot
```
nick1udwig: as on net, this should be a lot smaller. i'm imagining something like 10-20

kinode/src/vfs.rs
```diff
@@ -1030,21 +1017,25 @@ fn join_paths_safely(base: &PathBuf, extension: &str) -> PathBuf {
...
-                .close_least_recently_used_files(fraction_to_close)
-                .await?;
+        FdManagerRequest::FdsLimit(fds_limit) => {
+            files.fds_limit = fds_limit;
+            if files.open_files.len() as u64 >= fds_limit {
+                crate::fd_manager::send_fd_manager_hit_fds_limit(&files.our, &files.send_to_loop)
+                    .await;
+                files
+                    .close_least_recently_used_files(
+                        (files.open_files.len() as u64 - fds_limit) / 2,
```
nick1udwig: this isnt quite right: I think it should just be the difference without a denominator

nick1udwig (approved):
added a couple comments, otherwise lgtm. Thanks for fixing!

kinode/src/net/types.rs
dr-frmr: my thinking:

opening a new file has <1ms overhead, therefore, culling should happen rarely so as not to 2x or 3x overhead

connecting to a new peer has >10ms overhead due to network latency, therefore, culling old peer can happen every time without adding measurable overhead

nick1udwig: seems plausible, thanks for explaining

PR #576, opened 2024-10-16
Add contacts system userspace primitive

dr-frmr:
## Problem

Many userspace programs need to coordinate around "contacts", a set of other Kinode IDs that the node owner has some relationship to. Social programs in Kinode need to bootstrap their networks by pulling in an initial peer-set. This pattern exists in web2 social apps and is considered a must-have feature by many developers.

## Solution

Add `contacts:contacts:sys`, a userspace app that ships with the default Kinode distro and allows other userspace programs to get and set contacts. A contact is simply a KNS identity with optional 'fields', JSON objects mapped by a string label key. The contacts app serves 4 capabilities allowing users to granularly control access from other apps:
- read-name-only
- read
- add
- remove

The contacts app serves a basic frontend that allows Kinode users to directly manage and view their contacts, but most interaction with this program will likely be via API between it and other apps the user installs. Many apps will simply want `read-name-only` capability in order to bootstrap themselves with a set of nodes that the user may wish to engage with on the new app. 

## Testing

```
# Follow these steps to test
```

## Docs Update

[Corresponding docs PR](https://github.com/kinode-dao/kinode-book/pull/254)


kinode/packages/contacts/pkg/ui/index.html
```diff
@@ -0,0 +1,146 @@
...
+                <path
+                    d="M280.37 148.26L96 300.11V464a16 16 0 0 0 16 16l112.06-.29a16 16 0 0 0 15.92-16V368a16 16 0 0 1 16-16h64a16 16 0 0 1 16 16v95.64a16 16 0 0 0 16 16.05L464 480a16 16 0 0 0 16-16V300L295.67 148.26a12.19 12.19 0 0 0-15.3 0zM571.6 251.47L488 182.56V44.05a12 12 0 0 0-12-12h-56a12 12 0 0 0-12 12v72.61L318.47 43a48 48 0 0 0-61 0L4.34 251.47a12 12 0 0 0-1.6 16.9l25.5 31A12 12 0 0 0 45.15 301l235.22-193.74a12.19 12.19 0 0 1 15.3 0L530.9 301a12 12 0 0 0 16.9-1.6l25.5-31a12 12 0 0 0-1.7-16.93z">
+                </path>
+            </svg></button>
+        <h1>contacts</h1>
+    </span>
+    <main>
+        <article id="edit">
+            <form id="add-contact">
+                <input type="text" name="node">
```
nick1udwig: Perhaps add `placeholder = "Node name (e.g. my-friends-node.os)"` or something similar here? Else it is not entirely clear what you should type here

kinode/packages/contacts/pkg/ui/script.js
```diff
@@ -0,0 +1,126 @@
...
+                <button class="remove-field" onclick="removeField('${node}', '${field}')">X</button>
+            </li>
+        `).join('')}
+        </ul>
+        <form class="delete-contact" id="${node}">
+            <button type="submit">delete</button>
+        </form>
+        <form class="add-field" id="${node}">
+            <input type="text" name="field" placeholder="Field">
+            <input type="text" name="value" placeholder="Value">
```
nick1udwig: Consider adding an example to the placeholders, e.g.:

```
            <input type="text" name="field" placeholder="Field (e.g. Name)">
            <input type="text" name="value" placeholder="Value (e.g. Bob Ross)">
```

kinode/packages/contacts/api/contacts:sys-v0.wit
```diff
@@ -0,0 +1,36 @@
+interface contacts {
+    enum capabilities {
```
nick1udwig: nit: I think this should probably be singular, e.g., consider usage

```
Capability::Read
```

kinode/packages/contacts/contacts/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+            ),
+        );
+    }
+}
+
+wit_bindgen::generate!({
+    path: "target/wit",
+    world: "contacts-sys-v0",
+    generate_unused_types: true,
+    additional_derives: [PartialEq, serde::Deserialize, serde::Serialize],
```
nick1udwig: Consider using `process_macros::SerdeJsonInto` which will reduce `serde_json` usage below. I'd be happy to do a pass through and convert over to using if you'd like

nick1udwig (changes requested):
A few nits & UI recommendations. In addition, there's a problem with our "Home" button in the top-left corner. When clicked, it goes to the secure-subdomain-prefixed root page, rather than the root page, i.e. `http://contacts-sys.localhost:8080/` instead of `http://localhost:8080/`

kinode/packages/contacts/contacts/src/lib.rs
dr-frmr: good idea, i'll do it to get in the habit

PR #577, opened 2024-10-16
Improve login redirect

dr-frmr:
## Problem

Logic being used in login redirect from node registration FE -> node FE was faulty, particularly on windows. 

## Solution

Make it more robust by polling an endpoint that does not exist on the registration FE to trigger redirect

## Testing

```
Log in to a node
```

## Docs Update

N/A


nick1udwig (approved):
:rocket:

PR #586, opened 2024-10-28
Reduce log spam in case of inexistent processes being messaged

jaxs-ribs:
## Problem

When an old process that doesn't exist anymore is being messaged by other processes, a line is printed in the terminal. With bigger applications, this results in >50 lines printed per second. Malicious nodes can also log spam the terminal of someone else that way. 

## Solution

Setting the log level of the message to 2.


dr-frmr (approved):
Good change

Issue #584, opened 2024-10-24
feature: UI uninstall an unlisted application


nick1udwig:
https://discord.com/channels/1186394868336038080/1186421052071477428/1299043993207836695

PR #587, opened 2024-10-28
app_store UI: enable uninstalling local (unlisted) packages

barraguda:
## Problem

https://github.com/kinode-dao/kinode/issues/584

If an app was unlisted, there's currently no way for a user to uninstall an app (except using the uninstall script from the terminal)

## Solution

Adds a list of installed packages (disabling uninstall of core packages) in /my-apps (previously /my-downloads)

Still includes the downloaded zips and actions to install/delete those.

## Notes

Our basic vite configs work fine with subdomains, only trick is to go to 

```http://app-store-sys.localhost:3000/main:app_store:sys```

instead of 

```http://localhost:3000/main:app_store:sys```

might be a vite thing I can tweak so it doesn't throw you to the wrong url by default

### DIFF
--- a/kinode/packages/app_store/ui/src/App.tsx
+++ b/kinode/packages/app_store/ui/src/App.tsx
@@ -2,13 +2,13 @@ import React from "react";
 import { BrowserRouter as Router, Route, Routes } from "react-router-dom";
 
 import Header from "./components/Header";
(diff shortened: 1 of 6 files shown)


dr-frmr (approved):
Gets the job done, but we gotta get the UI looking more polished here lol

Issue #527, opened 2024-09-04
feature: gracefully handle `-`s vs `_`s in wasm filenames


nick1udwig:
**Is your feature request related to a problem? Please describe.**
https://github.com/kinode-dao/kit/issues/216 documents annoyance of cargo preferring `_`s and Kimap App Store protocol preferring `-`s. https://github.com/kinode-dao/kit/pull/222 resolved this on the `kit` side by making `kit` more robust wrt `_`s vs `-`s and just "doing the right thing". There is still some trickiness here associated with our `manifest.json` files, where we specify a file path for processes.

**Describe the solution you'd like**
Make Kinode core, when given a process path of `foo-bar.wasm`, if not found, check for `foo_bar.wasm` (and vice versa!). Print a verbosity 1 message if this switcheroo happens so that devs can be informed.

Ideally this should happen for both `manifest.json`, `scripts.json`, and `spawn()` to give a consistent behavior across all process-spawning interfaces.

**Describe alternatives you've considered**
This is in addition to `kit` being better. E.g., `kit` could read `manifest.json` and see if the corresponding `.wasm` file exists; error out with useful error message if it does not.

**Additional context**
@dr-frmr what do you think about this?

dr-frmr:
Yup, ought to do this

dr-frmr:
Revisiting this @nick1udwig I am now thinking that we should strictly enforce non-usage of underscore. Otherwise, every program that wants to interact with .wasm files stored in the VFS will always need to implement this backup-check logic. It seems burdensome as opposed to following the config files blindly and immediately alerting the developer if they are misconfigured.

nick1udwig:
Hmm, yeah, its either every `.wasm` interaction has the backup logic or else fail out ASAP.

I can change `kit` to error out given `_`s. This would be a breaking change.

nick1udwig:
https://github.com/kinode-dao/kit/pull/257 implements the desired "fail fast" behavior in `kit`.

At a minimum we should change distro modules/processes to use `-`s rather than `_`s. This would be a breaking change unless we alias them.

We should also consider rejecting running `.wasm` files that have a `_` in them. This would be a breaking change as well.

dr-frmr:
Is this work done / can we close this? @nick1udwig

nick1udwig:
Yup -- just waiting on 0.10.0 to merge, but work is done

PR #589, opened 2024-10-29
change `_`s to `-`s in distro modules/processes

nick1udwig:
## Problem

#527

We have restrictions on characters allowed in some fields in addresses (i.e. only a-z, A-Z, 0-9, `-`). It is confusing that these are not enforced/modeled.

## Solution

Enforce & make uniform across fields.

## Testing

TODO

## Docs Update

TODO

## Notes

BREAKING CHANGE: we'll have to test carefully to try to avoid breaking things for end-users, but we'll definitely break things for devs.

TODOs:
- [x] Restart throttling


dr-frmr:
Phat diff

dr-frmr:
I think we .gitignore the built app store JS

nick1udwig:
> I think we .gitignore the built app store JS

Ah, good call, didn't think about updating .gitignore entries

PR #590, opened 2024-10-30
fakenet: new addresses

barraguda:
## Problem

Anvil breaking change + change in contracts == new addresses.

## Solution

New addresses, kit PR: 

https://github.com/kinode-dao/kit/pull/260


dr-frmr:
@bitful-pannul this needs to go in 0.10 because it's breaking, right?

Issue #581, opened 2024-10-22
feature: capabilities devex


nick1udwig:
**Is your feature request related to a problem? Please describe.**
Its fiddly to get capabilities to work properly

**Describe the solution you'd like**
Two things that could be improved to make them less fiddly:
1. Reduce number of places they must be declared/specified. Currently need to specify in manifest.json/scripts.json as well as in the file itself, where they are attached to the message.
2. Improve visibility when debugging. Currently, parsing them in the event loop is possible, but not easy or good devex.

Some suggestions of some possible ways work towards these to get creative juices flowing:
1. Make it easy to attach all caps held by our process to a message. Or perhaps: all caps we have that are issued by the `target` process. The first possibility leads to a risk of leaking caps to nefarious third parties (since they can `save` them); so perhaps we'd need a concept of a "read only" capability. The second (auto attach all caps for a `target`) wouldn't have this problem. The downside here is that we are sending more information with each request. If you limit to local only, this might not be a huge problem, but still will lead to a performance hit.
2. Add some way to print attached caps with messages in event loop.

nick1udwig:
Realized 2 already exists. 1 might be a nice `process_lib` function.

nick1udwig:
Resolved in https://github.com/kinode-dao/process_lib/pull/120 https://github.com/kinode-dao/kinode/pull/619 ; will close issue once those are merged.

PR #596, opened 2024-11-05
terminal: m caps

### LINKED ISSUE #579: feature: add caps to `m` from terminal
**Is your feature request related to a problem? Please describe.**
Can't use `m` to make requests that require an attached capability.

**Describe the solution you'd like**
We should allow the user to specify the capability to attach to the request.

Perhaps automagically attach all caps that were issued by `foo:bar:baz.os` when sending a request to `foo:bar:baz.os`. I don't think we currently expose this information to processes, however.

nick1udwig:
## Problem

`m` doesn't send caps with messages, #579 #581

## Solution

Attach caps that were issued by a given process and that are held by terminal to `m` Requests.

## Testing

```
m our@contacts:contacts:sys "GetNames"
```

## Docs Update

None

## Notes

This still doesn't resolve #579 & #581 completely. The issue that remains is that there is no way for a process to pass a userspace cap to terminal. We either need to:
1. Change manifest field `grant_capabilities` to allow non-`messaging` caps to be granted,
2. Add a `terminal:terminal:sys` api that allows processes to give it capabilities.


dr-frmr:
> Change manifest field grant_capabilities to allow non-messaging caps to be granted

it's an oversight that this doesn't currently work -- i'm surprised this code does not allow it:
https://github.com/kinode-dao/kinode/blob/28664d7c27261dfd74127a85a020fd5452faa75e/kinode/packages/app-store/app-store/src/utils.rs#L342-L391

kinode/packages/terminal/pkg/manifest.json
```diff
@@ -7,6 +7,22 @@
...
+                "params": "Read"
+            },
+            {
+                "process": "contacts:contacts:sys",
+                "params": "Add"
+            },
+            {
+                "process": "contacts:contacts:sys",
+                "params": "Remove"
+            },
```
dr-frmr: Rather than requesting these in terminal (special case), let's do the general case thing and *grant* them to terminal:terminal:sys in the manifest for contacts:contacts:sys

dr-frmr (changes requested):
Just one change then can be merged I suppose

kinode/packages/terminal/pkg/manifest.json
nick1udwig: Agreed

nick1udwig: Sorry it took so long; changed in 108aacf

PR #601, opened 2024-11-09
v0.9.9

nick1udwig:


scripts/build-release.py
```diff
@@ -60,7 +60,7 @@ def build_and_move(feature, tmp_dir, architecture, os_name):
     source_path = f"target/release/{binary_name}"
     dest_path = os.path.join(tmp_dir, binary_name)
     shutil.move(source_path, dest_path)
-    os.chmod(dest_path, 0o775)
+    os.chmod(dest_path, 0o644)
```
nick1udwig: We may not need 775, but we at least need 744; 644 is not executable

dr-frmr: sorry, was blindly following this: https://github.com/Enigma-Dark/kinode-engagement/issues/4

i'll change to 744

PR #611, opened 2024-11-20
kernel: dont replace `last_blob` with None

nick1udwig:
## Problem

Doing Request/Response under-the-hood like in process_lib can lead to confusing side-effects, like killing the `lazy_load_blob`.

## Solution

Keep around the `lazy_load_blob` instead of replacing it with `None`. It is only replaced when a Message is received with a non-`None` `lazy_load_blob`.

## Testing

TODO

## Docs Update

TODO

## Notes

None


nick1udwig:
FYI @dr-frmr

PR #613, opened 2024-11-21
feat: improve homepage:sys protocol to add capabilities for setting system stylesheet and removing other apps' homepage icons

dr-frmr:
## Problem

Currently, uninstalling an app will not affect it's homepage icon or widget (other than breaking them!) until the node is rebooted. This is caused by apps only being able to add themselves -- killing an app does not have any effect on homepage because only an app could notify the homepage to add/remove itself.

## Solution

Add a feature to the `homepage:homepage:sys` protocol: `RemoveOther`, allowing any process with the `RemoveOther` capability to remove the icon+widget+metadata of *any* app on homepage.

This capability is requested/held by `main:app_store:sys` for obvious reasons. We can and should add a warning to the app store UI if any other app tries to request this capability.

Separately, we improve the `SetStylesheet` feature by adding a capability for this, too, and requesting it in `setttings:settings:sys`. Another one we should warn about.

## Testing

```
1. Add app to homepage by installing
2. Uninstall, observe that icon/widget is removed (takes a few seconds on debug build)
```

## Docs Update

N/A

## Notes

I encourage anyone interested in Kinode to review this PR's code diff, it gives a very clear summary of how to use capabilities in userspace. It also is instructive for core developers on how to make this exact code more ergonomic. But this entire PR took less than an hour.


barraguda (approved):
sexy!

PR #614, opened 2024-12-02
kns & app_store: store logs in state

barraguda:
## Problem

Re-indexing of both app_store and kns kimap logs is becoming bulky enough to soon warrant better indexing. 

## Solution

Save old verified logs in state, looking for new ones. 

Also, save state in extendable versioned enum way, so we can easily add/migrate in backwards compatible ways.

## Notes

- need to figure out HashMap<PackageId> key serializing (non strings x json not good)
- re-index button settings for kns [ ]
- re-index advanced settings button for app_store [ ]

### DIFF
--- a/kinode/packages/kns_indexer/kns_indexer/src/lib.rs
+++ b/kinode/packages/kns_indexer/kns_indexer/src/lib.rs
@@ -4,8 +4,8 @@ use crate::kinode::process::kns_indexer::{
 use alloy_primitives::keccak256;
 use alloy_sol_types::SolEvent;
 use kinode_process_lib::{
-    await_message, call_init, eth, kimap, net, print_to_terminal, println, timer, Address, Message,
-    Request, Response,
+    await_message, call_init, eth, get_state, kimap, net, print_to_terminal, println, set_state,
+    timer, Address, Message, Request, Response,
 };
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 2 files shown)


dr-frmr:
`app_store:sys: error saving state: key must be a string`

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -75,26 +168,27 @@ fn init(our: Address) {
...
 }
 
-fn main(our: Address, mut state: State) -> anyhow::Result<()> {
+fn main(our: Address, mut versioned_state: VersionedState) -> anyhow::Result<()> {
+    let mut state = versioned_state.state_mut();
+
     #[cfg(feature = "simulation-mode")]
     add_temp_hardcoded_tlzs(&mut state);
 
+    let contract_address = KIMAP_ADDRESS.parse::<eth::Address>().unwrap();
```
dr-frmr, line 186: There is a reason contract address was a field inside state. If the contract address changes, state must be reset. This was particularly an issue in beta as contract address could change -- extremely unlikely to happen on mainnet, but should still keep it in state especially now that we can perform state migrations in this app.

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -108,21 +202,22 @@ fn main(our: Address, mut state: State) -> anyhow::Result<()> {
...
+        .address(contract_address)
+        .from_block(state.last_block - 2)
         .to_block(eth::BlockNumberOrTag::Latest)
         .event("Note(bytes32,bytes32,bytes,bytes,bytes)")
         .topic3(notes);
 
     // 60s timeout -- these calls can take a long time
     // if they do time out, we try them again
-    let eth_provider: eth::Provider = eth::Provider::new(state.chain_id, SUBSCRIPTION_TIMEOUT);
+    let eth_provider: eth::Provider = eth::Provider::new(CHAIN_ID, SUBSCRIPTION_TIMEOUT);
```
dr-frmr, line 213: Chain ID should also be in state so it can migrate smoothly

dr-frmr: Imagine a future where kns_indexer reads from multiple chains. Not too far away

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -52,12 +64,93 @@ struct State {
...
+    }
+}
+
 // note: not defined in wit api right now like IndexerRequests.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 enum IndexerResponses {
     Name(Option<String>),
     NodeInfo(Option<net::KnsUpdate>),
-    GetState(State),
+    GetState(KnsState),
```
dr-frmr, line 153: This change needs to be reflected in process_lib unfortunately...

dr-frmr: Nevermind. Process_lib doesn't actually use this part of the enum.

barraguda:
superceded and addressed by #620

PR #615, opened 2024-12-03
remove ws register encrypted field

nick1udwig:
## Problem

Remove the `WsRegister::encrypted` field.

## Solution

^

## Testing

Not sure where this is used?

## Docs Update

TODO

## Notes

None


nick1udwig:
> Not sure where this is used?

It looks like our JS client lib doesn't send this anymore: https://github.com/kinode-dao/client-encryptor/commit/3155500b8a519ba6331ed5e18ce0f58f520d6102

Though our docs indicate it should be sent: https://github.com/kinode-dao/kinode-book/blob/2f70f9a9f0caccdd8d487819d4379fdccc6eb2be/src/apis/websocket.md?plain=1#L4

nick1udwig:
Seems to work, but secure subdomains not working on fake node. I suspect that was broken elsewhere, but I'll leave this open till we confirm.

nick1udwig:
Yup, seems to be working fine according to my tests -- thanks for the upstream fix

PR #617, opened 2024-12-03
terminal: per process verbosity

nick1udwig:
## Problem

It'd be nice to be able to control not just the overall verbosity of the Kinode terminal, but also the verbosity of individual processes.

## Solution

- [x] Add a `source: ProcessId` field to `Printout` (4271525)
- [x] Add `terminal` state of per-process-verbosity: `HashMap<ProcessId, u8>` (70a2d68)
- [x] Use that state to decide whether to print or not (70a2d68)
- [ ] ~~Add scripts to manipulate that state (e.g. `process-verbosity <ProcessId> <u8>` and perhaps `package-verbosity <PackageId> <u8>`)~~
- [x] Add "process verbosity mode" to allow setting per-process verbosity (aca7cdb)
- [x] Add runtime argument to set pre-process verbosities at boot (b5dae0c)

## Testing

TODO

## Docs Update

TODO

## Notes

[Context](https://discord.com/channels/1186394868336038080/1186684706700410962/1307097791864111124)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3965,7 +3965,7 @@ dependencies = [
 [[package]]
 name = "kinode_process_lib"
 version = "0.10.0"
-source = "git+https://github.com/kinode-dao/process_lib?rev=5d6e2af#5d6e2af88ef0716822cc1638be5b71a1ec9c4fb3"
+source = "git+https://github.com/kinode-dao/process_lib?rev=8a5b040#8a5b040dd699ae69d5bc1826fdd6e0bd5d84e766"
 dependencies = [
  "alloy 0.1.4",
  "alloy-primitives 0.7.7",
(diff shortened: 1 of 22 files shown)


nick1udwig:
Thinking about [this line](https://github.com/kinode-dao/kinode-wit/pull/9/files#diff-1fdd318d5c9b2afbdf5a39e4db4fb6b6d03bd61056e95bd21ca1d55d2ac46276R117) and whether we should be able to "mute" processes that are annoying makes me think it might be nice to have a "muted" option for individual processes. Any thoughts?

dr-frmr:
> Thinking about [this line](https://github.com/kinode-dao/kinode-wit/pull/9/files#diff-1fdd318d5c9b2afbdf5a39e4db4fb6b6d03bd61056e95bd21ca1d55d2ac46276R117) and whether we should be able to "mute" processes that are annoying makes me think it might be nice to have a "muted" option for individual processes. Any thoughts?

I don't see how it relates to `spawn`, but yes, muting a process would be smart. Like setting verbosity to "-1" to not even show 0-level prints, lol. Too bad we started at 0, otherwise could just use that.

EDIT: I see the exact line linked now. Yeah

nick1udwig:
Apologies for the bad link. Okay cool I'll add muting :+1:

PR #620, opened 2024-12-09
kns & app_store: new serialization strategy

barraguda:
## Problem

Versions of kns_indexer that keep hashmaps of names in memory (and or serialize that entire state and save it) will at some point run into memory limits and or just performance issues when indexing many events. 

Same goes for app_store.

## Solution

This PR uses key value as a hashmap on disk for KNS, doing gets and sets directly to disk.

For app_store, it uses the sqlite runtime module for better querying capabilities in the future too,

## Notes

Some things that will still be added:

- iterator support for KvAction (needed for state printouts now that we don't keep all keys in memory)
  - this is also needed for the consequent update of app_store (serves GetApps request often) 
- reset logic (now involving a full delete of the package db in /kv inside the home folder)


barraguda:
note: for the app_store chain module, which often gets the request GetApps, we need either an iterator function in kv to gather all the keys, or we keep some in memory. 

we need the iterator types anyway so I'll add them in a separate PR!

dr-frmr:
Does this fully replace #614 ?

barraguda:
> Does this fully replace #614 ?

yes, app_store work pending

reviews from that one have been assessed here though

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -41,15 +43,202 @@ const DELAY_MS: u64 = 1_000; // 1s
...
+        state.version = state.get_version();
+        state.last_block = state.get_last_block();
+
+        println!(
+            "kns_indexer: loaded state: version: {}, last_block: {}, chain_id: {}, kimap_address: {}",
+            state.version,
+            state.last_block,
+            state.get_chain_id(),
+            state.get_contract_address()
+        );
```
dr-frmr: if this is going to print on every boot it should be aesthetically pleasing

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -70,18 +259,10 @@ call_init!(init);
...
-        chain_id: CHAIN_ID,
-        contract_address: KIMAP_ADDRESS.parse::<eth::Address>().unwrap(),
-        nodes: HashMap::new(),
-        names: HashMap::new(),
-        last_block: KIMAP_FIRST_BLOCK,
-    };
+    // state is loaded from kv, and updated with the current block number and version.
+    let state = State::load(&our);
+
+    println!("got last block: {}", state.get_last_block());
```
dr-frmr: remove?

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -108,29 +290,31 @@ fn main(our: Address, mut state: State) -> anyhow::Result<()> {
 
     // sub_id: 2
     let notes_filter = eth::Filter::new()
-        .address(state.contract_address)
+        .address(state.get_contract_address())
```
dr-frmr: optimization: get contract address once at beginning and use it multiple times? calling kv.get() over and over is a lot, no?

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -108,29 +290,31 @@ fn main(our: Address, mut state: State) -> anyhow::Result<()> {
...
+        eth::Provider::new(state.get_chain_id(), SUBSCRIPTION_TIMEOUT);
 
     print_to_terminal(
         1,
         &format!(
             "subscribing, state.block: {}, chain_id: {}",
-            state.last_block - 1,
-            state.chain_id
+            state.get_last_block() - 1,
+            state.get_chain_id()
```
dr-frmr: same thing here with repeated calls to get_last_block, get_chain_id

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
```diff
@@ -41,15 +43,202 @@ const DELAY_MS: u64 = 1_000; // 1s
...
-    nodes: HashMap<String, net::KnsUpdate>,
+    // version of the state in kv
+    version: u32,
     // last block we have an update from
     last_block: u64,
+    // kv handle
+    // includes keys and values for:
+    // "meta:chain_id", "meta:version", "meta:last_block", "meta:contract_address",
+    // "names:{namehash}" -> "{name}", "nodes:{name}" -> "{node_info}"
+    kv: Kv<String, Vec<u8>>, // todo: maybe serialize directly into known enum of possible types?
```
dr-frmr: can you resolve this TODO before merge? what is the type going in here? it looks like just nodes, and then a handful of static types

dr-frmr (changes requested):

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
barraguda: something like this maybe? 
```
Tue 17:09 kns-indexer:sys:
     🐦‍⬛  KNS Indexer State
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
Version         1
Last Block 129424597
Chain ID       10
KIMAP      0xcA92476B2483aBD5D82AEBF0b56701Bb2e9be658
▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
```

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
barraguda: it's only called during the initial subscribing process, never after actually entering the main loop, I think it's okay to keep it that way. It's passed identically to 2 filters and a provider though, I can simplify that with a memory variable

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
barraguda: yep, the KV struct currently takes in a type for the value, I'll see if I can relax that requirement in the process_lib with a specific function passing in <K,V> rather than at instantiation. 

non-static-types right now: Strings for names and namehashes, net::KnsUpdates for node_infos

kinode/packages/kns-indexer/kns-indexer/src/lib.rs
```diff
@@ -42,54 +44,228 @@ const DELAY_MS: u64 = 1_000; // 1s
...
+            state.version,
+            state.last_block,
+            state.get_chain_id(),
+            state.get_contract_address().to_string()
+        );
+
+        state
+    }
+
+    fn meta_version_key() -> &'static str {
```
nick1udwig: Is it best practice to return a static str and then `.to_string()` it, as we do below?

nick1udwig (approved):
small question otherwise LGTM

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
dr-frmr: line up the number values horizontally but yes :)

kinode/packages/kns_indexer/kns_indexer/src/lib.rs
barraguda: using set_as::<u64> and get_as::<u64> reduced code amount here some, but left the default as Vec<u8> instead of having some bigger enum to match and serialize.

https://github.com/kinode-dao/process_lib/pull/123

kinode/packages/kns-indexer/kns-indexer/src/lib.rs
barraguda: probably not, kv helper takes ownership to serialize, so made those helpers just return a String directly instead!

kinode/packages/kns-indexer/kns-indexer/src/lib.rs
```diff
@@ -254,22 +361,19 @@ fn main(our: Address, mut state: State) -> anyhow::Result<()> {
                     // TODO: make sure we've seen the whole block, while actually
                     // sending a response to the proper place.
                     Response::new()
-                        .body(IndexerResponse::Name(state.names.get(hash).cloned()))
+                        .body(IndexerResponse::Name(state.get_name(hash)))
                         .send()?;
                 }
 
                 IndexerRequest::NodeInfo(NodeInfoRequest { ref name, .. }) => {
```
barraguda, line 368: note: should we add a kimap.get() call here (if we happened to miss an event), or have a separate IndexerRequest for it?

dr-frmr: yes we should. lazy eval/correctness in indexer. also in `net` runtime module need to do same thing.

PR #629, opened 2024-12-17
kv: iterator support

barraguda:
## Problem
https://github.com/kinode-dao/process_lib/pull/125

## Solution

```rust
pub enum KvAction {
    // ...
    IterStart { prefix: Option<Vec<u8>> },
    IterNext { iterator_id: u64, count: u64 },
    IterClose { iterator_id: u64 },
}

pub enum KvResponse {
    // ...
    IterStart { iterator_id: u64 },
    IterNext { done: bool },
    IterClose { iterator_id: u64 },
}
```
## Testing

```
# Follow these steps to test
```


nick1udwig (approved):
Seems reasonable to me. I'd recommend closing the iterator once you hit `done = true` rather than requiring the consumer to call `IterClose`.

In general `DashMap`s make me nervous cuz we've had a lot of deadlocks from them, but just looking at the code it seems ok :sweat_smile: 

Please be sure to hit all the new codepaths here in tests, ideally with two processes to try to exercise the `DashMap`s in a multi-tokio::task context.

lib/src/kv.rs
```diff
@@ -20,6 +20,10 @@ pub enum KvAction {
     BeginTx,
     Commit { tx_id: u64 },
     Backup,
+    // Iterator operations
```
dr-frmr: Comments inside enums should be `///` docstrings -- please label each iter operation with an explanation of how to use it. It's not immediately obvious to me how it works.

lib/src/kv.rs
```diff
@@ -28,6 +32,10 @@ pub enum KvResponse {
     BeginTx { tx_id: u64 },
     Get { key: Vec<u8> },
     Err { error: KvError },
+    // Iterator responses
```
dr-frmr: Same here re: comments

lib/src/kv.rs
```diff
@@ -38,6 +46,8 @@ pub enum KvError {
     KeyNotFound,
     #[error("no Tx found")]
     NoTx,
+    #[error("Iterator not found")]
```
dr-frmr, line 122: For all of the `KvError` strings here let's be more consistent in format (same capitalization).

dr-frmr (changes requested):
Comment nits, please try to give interface that final polish.

Echoing Nick's concern here about lockups when multiple processes are iterating on same DB, please make sure to test that.

PR #641, opened 2024-12-21
Dr/kv overhaul

dr-frmr:
please give feedback on kv types,
caps had big vuln, fixed
kv runtime was cloning unnecessarily so refactored some of that out


nick1udwig:
Looks good :+1: 

I wonder if there should be a DbKey type? I guess the choice to not have it is to make it easier for consumers of the API (since they won't have to, e.g., import that type in order to talk to kv)?

lib/src/kv.rs
```diff
@@ -1,79 +1,80 @@
...
     Backup,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum KvResponse {
     Ok,
     BeginTx { tx_id: u64 },
-    Get { key: Vec<u8> },
+    Get(Vec<u8>),
```
barraguda, line 102: now that it's no longer super obvious that the vec here is the original key used in the get request, we should at least add a comment that specifies that the returned value will be in the blob

lib/src/kv.rs
```diff
@@ -1,79 +1,80 @@
...
     pub db: String,
     pub action: KvAction,
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum KvAction {
     Open,
     RemoveDb,
     Set { key: Vec<u8>, tx_id: Option<u64> },
```
barraguda, line 47: same here, //value in blob

PR #642, opened 2024-12-21
Dr/sqlite overhaul

dr-frmr:


barraguda (approved):
changes look good, I'll test with some apps to see if caps behaviour and others remain unchanged!

lib/src/sqlite.rs
```diff
@@ -1,43 +1,41 @@
...
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum SqliteAction {
     Open,
     RemoveDb,
     Write {
         statement: String,
         tx_id: Option<u64>,
     },
```
barraguda, line 56: nit: same here as kv, should probably specify the type expected in the blob in a comment?

PR #653, opened 2024-12-27
app-store: remove hi

nick1udwig:
## Problem

We use `net:distro:sys` hi to test for node connectivity. This is bad practice because it requires `net` caps, which are very sensitive.

## Solution

Use a `Ping` variant.

## Testing

I wasn't able to test this out in the time I had today. Testing would look like:
1. Publish app on `foo.os`
2. Try to download the app on `bar.os` with event loop on for `main:app-store:sys`
3. Note the outgoing/incoming `Ping`s

## Docs Update

None

## Notes

None

### DIFF
--- a/kinode/packages/app-store/app-store/src/lib.rs
+++ b/kinode/packages/app-store/app-store/src/lib.rs
@@ -30,7 +30,7 @@
 //! It delegates these responsibilities to the downloads and chain processes respectively.
 //!
 use crate::kinode::process::downloads::{
-    AutoDownloadCompleteRequest, DownloadCompleteRequest, DownloadResponses, ProgressUpdate,
+    AutoDownloadCompleteRequest, DownloadCompleteRequest, DownloadResponse, ProgressUpdate,
 };
 use crate::kinode::process::main::{
     ApisResponse, GetApiResponse, InstallPackageRequest, InstallResponse, LocalRequest,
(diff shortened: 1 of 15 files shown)


dr-frmr:
@nick1udwig -- sending that raw bytes message to *another* node's net module does not actually require the net cap. Because it's a remote process, you can send it as long as you have networking capabilities, which app store does.

The actual reason that app-store has messaging to/from `net:distro:sys` is so that it can *give* that capability to other apps when they are installed. This is a historical artifact. It used to be the case that app-store gave capabilities from its own store to newly installed apps. Now, it just uses the kernel API directly. So we can actually get rid of a lot of the capabilities in app-store's list. Let's do that!

kinode/packages/app-store/app-store/src/lib.rs
```diff
@@ -131,6 +132,11 @@ fn handle_message(
                     response.send()?;
                 }
             }
+            Req::RemoteRequest(remote_request) => match remote_request {
+                RemoteRequest::Ping => {
+                    Response::new().body(RemoteResponse::Ping).send()?;
+                }
+            },
```
dr-frmr: Since the purpose of the ping is to determine whether or not a node is mirroring a given app, why don't we just go ahead and actually encode the requested package name into the request? @barraguda is there a reason why the app store doesn't have a *real* mirror-check remote request-response?

PR #663, opened 2025-01-08
app_store: fix launch button

barraguda:
## Problem

"app has no UI" displayed even when app had registered itself with homepage. 

/apps served by root homepage not accessable from subdomain.

## Solution

Add getApps to homepage wit types, add /homepageapps to app_store http, which then awaits data from homepage.

## Notes

Instead displaying "No UI for app" now just doesn't render any button at all. 

Also, download page to be refactored into a simpler modal


dr-frmr (approved):
LGTM. Non-breaking additive change to homepage API; can keep that at v1. Edited some comments but otherwise all set to merge.

PR #664, opened 2025-01-13
app store: rework FE

dr-frmr:
## Problem

App store had some issues, elaborated here:
https://gist.github.com/dr-frmr/5a3b40277a74bd17167a03a417a3d6fa

## Solution

Also elaborated here, change a number of FE components + refactor the widget. Mostly for user experience.
https://gist.github.com/dr-frmr/5a3b40277a74bd17167a03a417a3d6fa

## Testing

Use new app store!
- download and install apps
- publish and update apps
- test failures, edge cases

## Docs Update

N/A


barraguda:
When I go through the download/install flow everything is smooth, and launch/auto_update buttons show up: 
Screenshot 2025-01-15 at 23 19 36

But, upon refresh, they don't show anymore: 
Screenshot 2025-01-15 at 23 19 44


Probably a case of some useAppStore fetch hook not being called properly?

barraguda:
small note, previously I've enjoyed publishing/updating apps for node1.os from node2.os, which works if they're owned by the same eth address, this is no longer possible (might be perfectly reasonable)

Screenshot 2025-01-15 at 23 31 58

barraguda:
I've only seen this 2 times and am still trying to pinpoint where in the code it could happen, but upon a download finishing I have gotten "failed to install try again"
and in the console: `Installation flow failed: Error: Manifest not found for the selected version`

refresh fixed it, race condition from going too fast before manifest gets extracted?

barraguda:
Publishing erroring: `Error: The contract function "execute" reverted with the following reason:
Array index is out of bounds.`

investigating if there's an easy fix
edit: fix in [579239e](https://github.com/kinode-dao/kinode/pull/664/commits/579239e3cc5ce9964cf146009a82a6c1baf34f34)

kinode/packages/app-store/chain/src/lib.rs
```diff
@@ -487,12 +487,13 @@ fn handle_local_request(our: &Address, state: &mut State, req: ChainRequest) ->
             }
         }
         ChainRequest::Reset => {
-            state.db.reset(&our);
+            // state.db.reset(&our);
```
barraguda: nothing happens on pressing the reset button right now

dr-frmr: it re-indexes and updates the entries instead of deleting the entire DB. when you delete the entire DB, you lose your auto_update status. if you have a better way to preserve auto_update status while fully deleting DB we can switch to that.

dr-frmr:
I think I fixed the launch button issue and the 'manifest not found' issue. can you take a look @barraguda?

nick1udwig (commented):
possibly a big deal:
* reset store button is confusing (not clear what it does) and has no progress bar of any kind to indicate if resetting is "done" or not

nits:
* if scroll down in big icon app selection page, then select an app, we get put in a scrolled-down app screen, which excludes the header

kinode/packages/app-store/chain/src/lib.rs
barraguda: sounds reasonable, that break branch isn't hit right now though. Will tweak to see if return(true) passes through handle_local_request()

dr-frmr: Ah I see the issue. Fixing

barraguda:
> I think I fixed the launch button issue and the 'manifest not found' issue. can you take a look @barraguda?

this looks fixed to me! it still starts a download with offline mirrors, I'll see if I can tweak MirrorSelector and it's set packages

kinode/packages/app-store/chain/src/lib.rs
```diff
@@ -601,7 +597,10 @@ fn handle_eth_log(
     listing.tba = tba;
     listing.metadata_uri = metadata_uri;
     listing.metadata_hash = metadata_hash;
-    listing.metadata = metadata.clone();
+    listing.block = block_number;
+    if !startup {
+        listing.metadata = metadata.clone();
+    }
```
dr-frmr, line 603: Aha.

dr-frmr (commented):
Everything seems to work for me. Ready to merge when you are @barraguda

kinode/packages/app-store/chain/src/lib.rs
barraguda: yep...

barraguda (approved):
lgtm! incredibly nice work on simplifying the flow and UI!

PR #686, opened 2025-02-06
really import the KINO_ACCOUNT_UPGRADABLE_IMPL

0x70b1a5:
## Problem

Var not imported before use.

## Solution

Import it.

## Testing

```
Publish an app via the web UI.
```


nick1udwig:
Thank you for fix! :rocket:

nick1udwig (approved):

PR #712, opened 2025-03-17
added sign:sys package for signatures

polwex:
## Problem
Lack of a simple and secure interface for signing messages with the Hypermap KPI without exposing `net:sys` to userspace

## Solution
Made a simple process to sign messages and verify signatures
## Testing
TBD
```
# Follow these steps to test
```

## Docs Update
TBD
[Corresponding docs PR](https://github.com/hyperware-ai/hyperware-book/pull/my-pr-number)

## Notes

{Any other information useful for reviewers.}


hyperdrive/packages/signer/sign/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "sign"
+version = "0.1.0"
+edition = "2021"
+
+[features]
+simulation-mode = []
+
+[dependencies]
+anyhow = "1.0.97"
+hyperware_process_lib = "1.0.3"
```
nick1udwig: Bump to 1.0.4

hyperdrive/packages/signer/sign/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub site: String,
+    pub time: u64,
+    pub nonce: Option<String>,
+}
+#[derive(Debug, Serialize, Deserialize)]
+struct SignResponse {
+    pub body: SignRequest,
+    pub message: Vec<u8>,
+    pub signature: Vec<u8>,
+}
```
nick1udwig: Move these types into the `login:sys-v0.wit` file. This may require a slight change in the types (e.g. you can have an enum similar to the above but there is no `{ }` syntax; you need to use `( )` to supply args).

See, e.g., https://github.com/hyperware-ai/hyperdrive/blob/main/hyperdrive/packages/homepage/api/homepage%3Asys-v1.wit

See also this specific subsection, which defines a few conventions we use for package WIT APIs: https://book.hyperware.ai/cookbook/package_apis.html#exporting-an-api

In particular, the `interface` should have name corresponding to the process

hyperdrive/packages/signer/sign/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+        match msg {
+            Err(_send_error) => {
+                // ignore send errors, local-only process
+                continue;
+            }
+            Ok(Message::Request { body, .. }) => handle_request(&our, &body).unwrap_or_default(),
+            _ => continue, // ignore responses
+        }
+    }
+}
```
nick1udwig: Preference for an `initialize` function more of this form: https://github.com/hyperware-ai/kit/blob/master/src/new/templates/rust/no-ui/chat/chat/src/lib.rs#L84-L92

We should be logging in error cases (which should never happen, but if they do, we want a record of them)

hyperdrive/packages/signer/sign/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+
+fn handle_request(our: &Address, request_bytes: &[u8]) -> Result<()> {
+    let req = serde_json::from_slice::<SignerRequest>(request_bytes)?;
+    match req {
```
nick1udwig: A nice side benefit of moving types into WIT API is we can then do stuff like this `try_into()?`: https://github.com/hyperware-ai/kit/blob/master/src/new/templates/rust/no-ui/chat/chat/src/lib.rs#L29

Note you can always `#[derive(SerdeJsonInto)]` on your in-Rust types as well.

Unfortunately this #derive magic is only for serde_json so far, but we can easily add another derive for rmp_serde to process_macros: https://github.com/hyperware-ai/process_macros/blob/master/src/lib.rs

For now, let's just make sure all serde_json's that can be turned into try_intos or intos are try_intos or intos

hyperdrive/packages/signer/sign/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+fn handle_sign(our: &Address, req: SignRequest, request_bytes: &[u8]) -> Result<()> {
+    let body = rmp_serde::to_vec(&NetAction::Sign)?;
+    let res = Request::to(("our", "net", "distro", "sys"))
+        .blob_bytes(request_bytes)
+        .body(body)
+        .send_and_await_response(10)??;
+    let Ok(NetResponse::Signed) = rmp_serde::from_slice::<NetResponse>(res.body()) else {
+        return Err(anyhow!("signature failed"));
+    };
+    let newblob = res.blob();
```
nick1udwig: Make this var name more meaningful, maybe something like `signature`

hyperdrive/packages/signer/sign/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+    match newblob {
+        None => Err(anyhow!("no blob")),
+        Some(b) => {
+            let lr = SignResponse {
+                body: req,
+                message,
+                signature: b.bytes().to_vec(),
+            };
+            let lrj = serde_json::to_vec(&lr)?;
+            Response::new().body(lrj).send()?;
```
nick1udwig: Can we give `lr` and `lrj` more meaningful names?

nick1udwig (changes requested):
Nearly there. One major change: moving types from Rust to WIT. Otherwise just stylistic.

PR #724, opened 2025-03-28
lightmode and appstore

0x70b1a5:
## Problem

Lightmode links are messed up; appstore needs some "mobility"

## Solution

Continue to take the axe to all manner of inutile Cascading Style Sheet rules

## Testing

```
Open the app store in lightmode, 
click around, 
resize the browser, etc.... 
same on homepage.
```

## Docs Update
Nope

### DIFF
--- a/css/hyperware.css
+++ b/css/hyperware.css
@@ -27,6 +27,8 @@
     --primary-light: var(--neon-green-light);
     --primary-xlight: var(--neon-green-xlight);
     --secondary-color: var(--iris);
+    --secondary-light: var(--iris-light);
+    --secondary-xlight: var(--iris-xlight);
     --tertiary-color: var(--orange);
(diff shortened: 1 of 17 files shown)


nick1udwig:
Looks really good. Found two issues:

1. color issue on registration flow: the progress bar thing on light mode:

Screenshot from 2025-03-28 16-00-53

2. default icon in app store (i.e. if no icon supplied) is left-justified in mobile view:

image


After fixing those, merge away! Really nice work

We'll have to revisit mobile App Store, but I think not for this PR. Not sure if its work for us or for a designer, but it should look a lot more like the "state of the art" app stores (iOS and Google Play):

image

nick1udwig (approved):
Please make requested changes, then merge when ready

PR #740, opened 2025-05-28
fix: typos in function name

crStiv:
nothing fancy, just found a typos in function name and fixed it

## Before

check_reponse

## After

check_response


nick1udwig (approved):
Thanks for the fix! :rocket:

PR #785, opened 2025-07-16
test for keyfile in login_with_password

### LINKED ISSUE #775: bug: trying to make a new node with `--password` flag crashes
```
# ~/hyperdrives/1.4.0 ~/hyperware-homes/foo.os --password hunter2

...

thread 'main' panicked at hyperdrive/src/main.rs:899:10:
could not read keyfile: Os { code: 2, kind: NotFound, message: "No such file or directory" }
Aborted (core dumped)
```

this behavior is fine, but the error message should be more useful

dolled-possum:
## Problem

If hyperdrive is launched with the `--password` flag but no existing keyfile (such as when booting a new node), it panics. We can do better.

## Solution

Fixed by updating the login_with_password fn to test for the existence of the keyfile rather than expecting it. If not found, returns a message explaining the user cannot use --password without a keyfile, and where hyperdrive expected to find it given the user's specified node directory. If instead the file is found but is otherwise unreadable, display the file path and the Error. In either case, cleanly exit with a status of 1.

## Testing

# Follow these steps to test
`./hyperdrive my_not_yet_existing_node_directory --password sUMpASSWORD`
will return
```
Error: Cannot use --password flag because keyfile not found at expected location: /your_current_dir/my_not_yet_existing_node_directory/.keys
You need to specify a keyfile first or use the registration flow (without --password) instead.
```

### DIFF
--- a/hyperdrive/src/main.rs
+++ b/hyperdrive/src/main.rs
@@ -894,9 +894,21 @@ async fn login_with_password(
     use argon2::Argon2;
     use ring::signature::KeyPair;
 
-    let disk_keyfile: Vec<u8> = tokio::fs::read(home_directory_path.join(".keys"))
-        .await
-        .expect("could not read keyfile");
+    let keyfile_path = home_directory_path.join(".keys");
+    let disk_keyfile: Vec<u8> = match tokio::fs::read(&keyfile_path).await {
+        Ok(file) => file,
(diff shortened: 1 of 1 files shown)


nick1udwig (approved):
Looks & works great, nice work!

PR #793, opened 2025-07-19
Dw/rpc-flag-preference (improves processing of --rpc and --rpc-config flags at startup).

### LINKED ISSUE #737: bug: rpc vs rpc-config and public nodes
**Describe the bug**

If `--rpc` is not passed, we default to publicnode, but we don't require --rpc to be passed after the first time -- not to mentioned `--rpc-config`.

We should always use a non-default RPC if we have it

https://github.com/hyperware-ai/hyperdrive/blob/bd9e5bd5a51a521fff434727c63fcaee525c27d9/hyperdrive/src/main.rs#L275
https://github.com/hyperware-ai/hyperdrive/blob/36f10c40da371656992e32d9bac97fdc9234c717/hyperdrive/src/register.rs#L42
https://github.com/hyperware-ai/hyperdrive/blob/36f10c40da371656992e32d9bac97fdc9234c717/hyperdrive/src/register.rs#L250-L255

dolled-possum:
## Problem

Support for --rpc-config was added some months ago, but the connect_to_provider function in register.rs still only accepted a URL string from the --rpc flag, ignoring any existing configuration or --rpc-config specified providers

## Solution

The handling of these flags has now been rationalized (and also made consistent with runtime handling as well).  We start by recognizing that --rpc is just a special case (one provider, no auth) of what can be specified in --rpc-config.  So only one or the other should be specified (if any), not both.

eth_provider_config will be potentially modified by these flags, adding one or more ProviderConfigs at the top of the vector (after deduplicating them if needed: this logic was cribbed from the add_provider_config fn in eth/mod.rs, and is kept in a new "add_provider_to_config" helper function in main.rs), and preserving the order if multiple ProviderConfigs are specified in the --rpc-config file.

The calls to serve_register_fe and login_with_password now pass in this whole eth_provider_config, not merely an RPC URL string.  This propagates as required, eventually calling the new function connect_to_provider_from_config, which replaces the previous connect_to_provider function.  Now it is possible for an authenticated connection to serve register and login with password.  It will also attempt to use all of the saved/configured RPC providers, only falling back to the default if no others succeed.

## Testing

There is currently a ton of logging in here that is meant only for diagnostic verbosity, and MUST be removed before merging.  To test, try all sorts of combinations of valid and invalid --rpc and --rpc-config specification.  The logging will show exactly what is going on.

## Notes

There's also an addition to .gitignore that helps out with jetbrains IDEs.


dolled-possum:
REMINDER: DEBUG LOGGING MUST BE REMOVED BEFORE MERGING

Issue #729, opened 2025-04-15
feature: add rpc management from within node


nick1udwig:
Currently can only manage RPCs from outside of node (i.e. from commandline). Add RPC management tools from within the node (i.e. script to list RPCs, add RPCs)

Related: RPC priority

dolled-possum:
RemoveProvider doesn't give any feedback regarding whether or not the Provider was actually found in the list.  I can update it, but then it might also make sense to update it in the Settings app as well.

dolled-possum:
Also perhaps by default GetProviders should mask the auth section, and only a new function, GetProvidersUnmasked, should reveal it.  I'll have to check if anything else calling GetProviders is expecting to use that auth data.

dolled-possum:
> Also perhaps by default GetProviders should mask the auth section, and only a new function, GetProvidersUnmasked, should reveal it. I'll have to check if anything else calling GetProviders is expecting to use that auth data.

Nah, I'll just mask it in the display.  No need to overdo it.  The Settings app buries the display a few screens deep, so not a priority.

dolled-possum:
Adding node providers from within the add-provider script does not currently capture the public key, IPs, and ports like it should be.  Making an issue for this, and will disable the ability to add node providers for the moment.

nick1udwig:
Looks to me like this issue can be closed @dolled-possum ?

dolled-possum:
> Looks to me like this issue can be closed [@dolled-possum](https://github.com/dolled-possum) ?

Yup, closing.  I had thought linking the PR and closing that automatically closed it, but apparently not; I've closed a second issue similarly associated with that PR.

nick1udwig:
Yeah the auto-closing is cool but seems somewhat brittle. I've had best results when I have the text in the PR OP

```
resolves #f00
```

PR #795, opened 2025-07-21
Dw/rpc flag handler

### LINKED ISSUE #737: bug: rpc vs rpc-config and public nodes

dolled-possum:
## Problem

inconsistent handling of --rpc vs --rpc-config flags

## Solution

The handling of these flags has now been rationalized (and also made consistent with runtime handling as well). We start by recognizing that --rpc is just a special case (one provider, no auth) of what can be specified in --rpc-config. So only one or the other should be specified (if any), not both.

eth_provider_config will be potentially modified by these flags, adding ProviderConfigs at the top of the vector (after deduplicating them if needed), and preserving the order if multiple ProviderConfigs are specified in the --rpc-config file.

The calls to serve_register_fe and login_with_password now accept this whole eth_provider_config, not merely an RPC URL string. This propagates as required, eventually calling the new function connect_to_provider_from_config, which replaces the previous connect_to_provider function. Now it is possible for an authenticated connection to serve register and login with password. It will also attempt to use all of the saved/configured RPC providers, only falling back to the default if no others succeed.

## Notes

I'll remove tagged debug logging before merging

### DIFF
--- a/hyperdrive/src/main.rs
+++ b/hyperdrive/src/main.rs
@@ -6,7 +6,6 @@ use lib::types::core::{
     NetworkErrorSender, NodeRouting, PrintReceiver, PrintSender, ProcessId, ProcessVerbosity,
     Request, KERNEL_PROCESS_ID,
 };
-use lib::types::eth::RpcUrlConfigInput;
 #[cfg(feature = "simulation-mode")]
 use ring::{rand::SystemRandom, signature, signature::KeyPair};
 use std::collections::HashMap;
@@ -88,6 +87,12 @@ async fn main() {
     let rpc_config = matches.get_one::<String>("rpc-config").map(|p| {
         std::fs::canonicalize(&p).expect(&format!("specified rpc-config path {p} not found"))
     });
+
+    // Prevent using both --rpc and --rpc-config flags simultaneously
+    if rpc.is_some() && rpc_config.is_some() {
+        panic!("Cannot use both --rpc and --rpc-config flags simultaneously. Please use either --rpc for a single RPC URL or --rpc-config for multiple URLs and/or advanced configs (e.g., auth), but not both.");
+    }
+
     let password = matches.get_one::<String>("password");
 
     // logging mode is toggled at runtime by CTRL+L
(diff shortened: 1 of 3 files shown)


hyperdrive/src/main.rs
```diff
@@ -129,38 +134,114 @@ async fn main() {
...
-                    provider: lib::eth::NodeOrRpcUrl::RpcUrl { url, auth },
-                },
-            );
+        // TODO: Remove debug logging before merging
+        eprintln!(
+            "[DEBUG-AUTH] Processing --rpc-config flag with file: {}",
+            rpc_config.display()
+        );
+
+        match std::fs::read_to_string(&rpc_config) {
```
nick1udwig, line 170: `match` is a great construct, but style-wise it should be used sparingly when other control flow is available because of the double-indentation it causes. Especially when there are two main outcomes we care about, we have two good alternatives:

alternative 1:

```rust
if let Ok(contents) = std::fs::read_to_string(&rpc_config) {
    ...
}
...
```

or (if we care about the non-Ok case, but don't need to unpack it)

```rust
if let Ok(...) ... {
} else {
}
...
```

alternative 2 (my preference when it works (e.g. when you want to exit early in a non-Ok case):

```rust
let Ok(contents) = std::fs::read_to_string(&rpc_config) else {
    ...
};
...
```

hyperdrive/src/register.rs
```diff
@@ -245,38 +245,139 @@ pub async fn register(
...
+            );
         }
     }
 
+    // TODO: Remove debug logging before merging
+    eprintln!("[DEBUG-AUTH] ❌ All providers (configured + fallback) failed!");
+
     panic!(
-        "Error: runtime could not connect to Base ETH RPCs {rpc_urls:?}\n\
+        "Error: runtime could not connect to any ETH RPC providers\n\
```
nick1udwig, line 378: `Base` is significant here because we need the Base RPC to bootstrap our node. Would prefer this went back to `Base` from `any`

hyperdrive/src/register.rs
```diff
@@ -245,38 +245,139 @@ pub async fn register(
...
         This is necessary in order to verify node identity onchain.\n\
-        Please make sure you are using a valid WebSockets URL if using \
-        the --rpc flag, and you are connected to the internet."
+        Please check your configured providers and internet connection."
```
nick1udwig, line 380: We now have 2 or 3 ways users can configure RPC so idk that it makes sense to list them all here. Maybe we should link to docs about how to set RPC?

I suppose this means deciding on and maintaining a canonical discussion somewhere. Two options I can think of are

https://book.hyperware.ai/kit/boot-real-node.html?highlight=rpc#--rpc
https://github.com/hyperware-ai/hyperdrive?tab=readme-ov-file#configuring-the-eth-rpc-provider

but it would ultimately be nice to give users as helpful an error message as possible.

nick1udwig (approved):
Style comment and a few nits. Correct the nits if you agree -- then feel free to merge after removing debugging stuff and confirming it still builds. Thanks for the nice work :rocket:

dolled-possum:
All very good points!  Because they aren't critical, I'll merge this branch (which has pulled in Tobias' Z index work as well) and clean these up in the PR for Issue #729 which should be done later today.

hyperdrive/src/register.rs
dolled-possum: There doesn't seem to be anything in the original code that limits the connections to Base eth RPCs, and my code maintains the status quo in the that regard.  I did update the message, but perhaps I should be enforcing some additional checks for chain ID 8453.  I'll look at that after the current work is done.

hyperdrive/src/register.rs
dolled-possum: Created an issue to address this.

hyperdrive/src/main.rs
dolled-possum: Updated, and because there were nested matches it avoided the dreaded quadruple indentation.

PR #799, opened 2025-07-23
Dw/prov mgmt scripts

### LINKED ISSUE #729: feature: add rpc management from within node

### LINKED ISSUE #801: Add ability to fully specify a node provider from the add-provider script (or perhaps a dedicated add-node-provider script)
> Adding node providers from within the add-provider script does not currently capture the public key, IPs, and ports like it should be.  Making an issue for this, and will disable the ability to add node providers for the moment. 

 _Originally posted by @dolled-possum in [#729](https://github.com/hyperware-ai/hyperdrive/issues/729#issuecomment-3113773025)_

dolled-possum:
## Description

This PR adds three terminal scripts for managing the provider config: get-providers, add-provider, and remove-provider.  They are included in terminal help and their usage guidelines are included when arguments aren't supplied. Testing is easy because you can see the results of add- or remove- by looking at the results of get-providers afterward.


dolled-possum:
In the eth module, I updated EthConfigAction::AddProvider to remove any already configured instances of that node name or URL before re-adding:  This makes it consistent with how --rpc-config and --rpc work, as well as giving the terminal user an easy way to update things like auth credentials without accidentally making a second copy of the provider.  However, I noticed that the weird ordering issue seems to happen with AddProvider, rather than guaranteeing that the newly added provider is first in the list.  I'm creating an issue for this for me to investigate.

dolled-possum:
add-provider is now split into add-node-provider and add-rpcurl-provider to better handle their different mandatory fields and optional params.  Also, add-rpcurl-provider defaults to the Base chain ID (but is overridable).

dolled-possum:
As I'm looking into the code structure more deeply for hypermap binding cacher, I noticed the packages list in the top level Cargo.toml, where I have _not_ yet added my scripts to the [workspace] members array.  Is this essential?  My scripts seem to be working without it, so I'm not sure yet what function it serves, will investigate.

hyperdrive/packages/terminal/add-node-provider/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[features]
+simulation-mode = []
+
+[dependencies]
+hyperware_process_lib = "2.0.0"
+rmp-serde = "1.1.2"
+serde = { version = "1.0", features = ["derive"] }
+wit-bindgen = "0.42.1"
+serde_json = "1.0.140"
```
nick1udwig, line 13: prefer alphabetical order when there isn't some other logical ordering to use (here and elsewhere)

hyperdrive/packages/terminal/get-providers/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+    //// Parse as generic JSON and pretty-print
+    //match serde_json::from_slice::<Value>(&body) {
+    //    Ok(json_value) => {
+    //        serde_json::to_string_pretty(&json_value)
+    //            .unwrap_or_else(|_| "Failed to format JSON".to_string())
+    //    }
+    //    Err(e) => {
+    //        format!("Failed to parse response as JSON: {}\nRaw response: {}", e, String::from_utf8_lossy(&body))
+    //    }
+    //}
```
nick1udwig: remove old commented-out code

nick1udwig (approved):
A couple small comments -- merge when ready.

Nice work, sorry it took so long to review! :rocket:

nick1udwig:
> As I'm looking into the code structure more deeply for hypermap binding cacher, I noticed the packages list in the top level Cargo.toml, where I have _not_ yet added my scripts to the [workspace] members array. Is this essential? My scripts seem to be working without it, so I'm not sure yet what function it serves, will investigate.

While you're here fixing those, might as well add the scripts to the top-level workspace as well!

PR #803, opened 2025-07-30
public appstore

0x70b1a5:
Test it: `[node url]/main:app-store:sys/public`


nick1udwig (changes requested):
Really nice work! Please make the following changes, then I'll do the final wiring to make this a toggle-able thing:

- [x] remove `pkg/public-ui` and add it to .gitignore
- [x] hide the Technical Details behind a pop-down button
- [x] consider adding the Technical Details (behind popdown button) to the non-public version of the page
- [x] call-to-action should take the form of a Download/Install button v similar to the non-public version of the page, however:
- [x] change the call-to-action: offer a link to Valet and to https://book.hyperware.ai/getting_started/install.html ; should say something along the lines of "To install and use this app, [get a hosted node](https://valet.hyperware.ai) or [host a node yourself](https://book.hyperware.ai/getting_started/install.html)"

The biggest thing here is to get the "install"/"call-to-action" flow as simple and obvious as possible: go check out the Google Play store to get an intuition for what I'm talking about wrt the "Download/Install button as CTA"

The ideal future flow for this (beyond scope of this PR) is:
1. User opens apps.hyperware.ai
2. User sees an app they want
3. User clicks into that app, clicks "Download" or "Install"
4. User is prompted to "Sign in" (to valet) with an alternative option to "Learn how to host your own node"
5. User goes through sign in/register flow on valet
6. User is spit out into their node's app store on the same app page as they were on

For this PR we should at least get 2-5

image

image

nick1udwig:
Ah, one more thing. With the "toggle public on / off" I will want to have a toggle somewhere in the non-public app store (maybe we can hide it in `My Apps > Advanced`? idk, but it should not be on the beaten path). We can either add the toggle and I'll hook it up to backend or add backend logic and hook it up to toggle. Which is more productive?

0x70b1a5:
@nick1udwig donezo, nick-sama

nick1udwig:
Excellent, I'll take it from here :saluting_face:

nick1udwig (approved):
Great work @0x70b1a5 :rocket:

PR #812, opened 2025-08-12
Implement Gabi feedback

0x70b1a5:
## Problem

https://docs.google.com/presentation/d/1ksSf1m-LT6tH6GfJ0QO7MOFAhzvuegyX7HYp3UrtoqA/edit?slide=id.g372575dd330_0_73#slide=id.g372575dd330_0_73

## Solution

Do the needful

## Testing

```
Test 1: onboarding
1. first open of homepage should show onboarding modal
2. clicking Let's go should close the modal
3. the clock widget should have an explanatory overlay re widget customizability
4. refreshing the page should show the onboarding modal again
5. clicking Don't show again should hide the modal forever and also hide the widget overlay

Test 2: edit and support buttons
1. they should be icon only on the left of Search apps
2. edit has same fn as b4
3. support opens your email client

Test 3: searching apps
1. With the appstore modal open, search in the homepage searchbar for an app.
2. backspace the search text
3. the appstore widget should not be blank or invisible anymore
4. search again for "dart" or another REAL APP that you DO NOT have installed
5. verify you don't see dartfrog on your homepage
6. verify you DO see a popup text suggesting to search the appstore.
7. click the popup
8. verify the appstore appears, with an autocompleted search for the text you entered.

Test 4: docked apps
1. docked apps should now show their names

Test 5: right-side home button
1. open appstore
2. click Home on right side
3. you should see the appstore slide closed and take you back home
```


0x70b1a5:
@nick1udwig ready for review

nick1udwig (changes requested):
We need to make sure mobile is the *first* target. These changes break a number of things on mobile:

- [x] Initial splash screen is good, but the copy talks about a desktop. Either rewrite or get input from bzns about how to rewrite this to focus on *mobile first*
- [x] Splash screen modal "X" button is behind the right side bar on mobile
- [x] Remove the "mailto" button. I'm happy to re-add this as a "DM using Official Chat App" button in future, but email is weird when considering a mobile first experience
- [x] The "Edit" button looks good but when clicked the buttons overrun mobile screen (I'm looking at iPhone 12 Pro dimensions)
- [x] I like this "link to app store if search not found", but really what search should be doing is opening the "All Apps" pane and searching therein. If no app is found, display that link
- [x] I'm not opposed to the recents/home button on the right, but we need to keep in mind we are mobile first: these buttons are pretty small and hard to click on mobile. That's OK for now, though we should think if we can figure out how to improve this. The REAL problem is that the right side bar doesn't work and neither do the buttons. They must work AT LEAST as well as master (ideally better) before merge
- [x] Dock overruns mobile width (I see "recent apps" going off screen)
- [x] While we're here, when you click/tap anywhere in the "recent apps" view other than a button or an app, we should go home

nick1udwig (approved):
Really nice. I like the draggable action button and it works well. We'll ask for more opinions when we get this into staging (so its easy for people to compare).

Let's fix this before merge:

- [x] The draggable action button should remember where it was placed for the next session (can just use browser storage)

Unrelated nits that you can either fix here or make issues for later as your prefer:

- [x] When in "app switch" view (i.e. after pressing the draggable button) on mobile I'd like to be able to drag any app left/right to close the app (current behavior is I have to click the `X` in top-right corner to close an app; this would be an additional way to close an app)
- [x] When tap "Apps" on mobile, should NOT focus the Search entry box; on desktop is good; when clicking into Search from homepage is good; but when tap "Apps" on mobile, focusing Search entry box opens keyboard which is annoying
- [x] When in "Apps" view, clicking/tapping anywhere non-app/search/action-button should return to home (similar to in the "app switch" view tapping anywhere returns home)
- [x] Get rid of the "Close" button on "Apps" view
- [x] Get rid of the "Home"/"Close" buttons on "app switch" view
- [x] On desktop, focusing the "Search apps" button behavior is correct, but the visual experience is weird because the search bar jumps from ~middle of screen to ~right of screen. The ideal case is the search bar looks the same in the two views

nick1udwig:
Beautiful. Really nice work. Love the dot!

PR #820, opened 2025-08-14
implement basename option 1

### LINKED ISSUE #818: feature: public app-store don't require big endpoint
https://apps.hyperware.ai/ redirects to https://apps.hyperware.ai/main:app-store:sys/public

the ideal behavior is that we rewrite the endpoint so that `/` becomes `/main:...`

we can do the rewriting with caddy:

```
apps.hyperware.ai {
    # Only rewrite paths that are NOT ACME challenges or assets
    @notacme_notassets {
        not path /.well-known/acme-challenge/*
        not path /assets/*
	not path /our.js
    }
    
    # Rewrite assets to the correct path
    @assets {
        path /assets/*
    }
    @our {
	path /our.js
    }
    
    rewrite @our {uri}
    rewrite @assets /main:app-store:sys{uri}
    rewrite @notacme_notassets /main:app-store:sys/public{uri}
    reverse_proxy localhost:8080
}
```

however, the app store is blank and console reports

```
<Router basename="/main:app-store:sys/public"> is not able to match the URL "/" 
because it does not start with the basename, so the <Router> won't render anything.
```

claude seems to think we can have a custom router that will be happy with either `/` OR that endpoint; if so, lets do that. If not, can we figure out another solution?

https://claude.ai/share/83207231-cc7c-4a5b-a63f-bf84fee82a74

0x70b1a5:
Fix #818


nick1udwig (approved):
Shoot, forgot to get this merged

Issue #811, opened 2025-08-11
feature: links to other apps


nick1udwig:
We should be able to put links to other apps like
```
hd://main:app-store:sys/foo
```
or
```
/main:app-store:sys/foo
```
and have it Just Work

When in an iframe in homepage app should be able to just open that app (if it exists; if not? Prompt to go look in App Store for that app?)

What about in localhost case?

0x70b1a5:
Dovetails with #805. If we solve this one first, the other one follows. I came to the issues board to raise this problem actually - the appstore frontend would need to talk to the homepage frontend somehow in order to integrate with the iframe js structure nicely anyhow. 

It's probably going to end up looking like some kind of `navigator:sys` process, since processes are all going to have to communicate with the central homepage store to coordinate open, opening, and closing apps, unless I'm mistaken

nick1udwig:
What are your thoughts on this alternative proposal? @0x70b1a5 

# prompt

  look at hyperdrive/packages/homepage/ui/src and 
  hyperdrive/packages/homepage/ui/src/components/AndroidHomescreen

  applications are served inside of iframes in the homescreen. I want to be able to link from 
  application to application, opening a new iframe for the appropriate application. Assess if 
  this is possible purely in the frontend. Don't write code

  for more context, applications are served at `/process:pacakge:publisher` and can serve 
  endpoints under that such as `/process:package:publisher/foo`

  i want to allow links to `/process:package:publisher` to open a iframe with that app i.e 
  using openApp in hyperdrive/packages/homepage/ui/src/stores/navigationStore.ts

  is it possible? if so, how would you do it?

# claude response

● Based on my analysis of the codebase, yes, it is possible to enable cross-iframe navigation
  purely in the frontend. Here's my assessment:

  Current Architecture

  The homepage serves as a container that loads applications in iframes. Each app is served at
   /process:package:publisher with the ability to serve sub-endpoints. The openApp function in
   navigationStore.ts:68-194 already handles opening apps in iframes.

  How to Implement Cross-App Linking

  You can achieve this through postMessage API communication between iframes and the parent
  homepage. Here's the approach:

  1. Parent Window Message Handler

  Add a message listener in the homepage (parent window) that listens for navigation requests
  from iframes:
  ```
  // In AndroidHomescreen component or main app
  window.addEventListener('message', (event) => {
    if (event.data.type === 'OPEN_APP') {
      const { process, package_name, publisher, path } = event.data;
[...]

Issue #805, opened 2025-08-04
bug: App Store `Launch` button doesn't play nice with new homepage


nick1udwig:
When user clicks `Launch`, the app opens (good) but it doesn't open in a new "window"/iframe and instead "takes over" the App Store "window"/iframe (bad)

Make it so that clicking `Launch` opens a new iframe, just like clicking the app icon on the homepage would

0x70b1a5:
also, node homepage has too dark bg on lightmode.

nick1udwig:
done in #840

Issue #802, opened 2025-07-30
feature: app-store UI: the "Launch" and "Install" buttons on the store page should do that thing when clicked


nick1udwig:
**Is your feature request related to a problem? Please describe.**
We should strive to minimize required user clicks in App Store

**Describe the solution you'd like**
If a user clicks "Launch" or "Install" on store page, do that action, don't take them to the app page and do nothing (as we currently do).

**Describe alternatives you've considered**
None

**Additional context**

Currently clicking "Launch" or "Install" is the same as clicking the entire app itself. They should be distinct actions.

<img width="838" height="823" alt="Image" src="https://github.com/user-attachments/assets/8fc353bf-6a6e-49a8-b02f-d80cdf5de7f9" />

0x70b1a5:
Ok, fair. But please allow me to explain the reasoning here. 

I made this choice because of the complexity of the state of an appstore app. Currently we're only drilling down into the mirror and install and details (each of those is, IIRC a separate HTTP request, per app) views when you get to the app's details page. The return value of each of those calls impacts the info displayed in the ui: if no mirrors are online, or something, the button should look and act different if it's actually going to be an action button rather than the current pseudobutton.

We could a) implement ui for all of this, but then for each of these elements we're making multiple calls to the backend just to get state, or b) write a new endpoint that will return all of the info we need in a single request. I would like the latter of course. Do you think I should do that work? If the code for the requests that are currently made to the backend on the app details page can be combined and re-invoked in the process code in a single method, it might be simple enough for me.

0x70b1a5:
I can see "Launch" doing what it says, though.

nick1udwig:
This is good info, thanks for raising.

b sounds the most promising; if you've got bandwidth for it, definitely see how far you can get. I think changes should largely be in [here](https://github.com/hyperware-ai/hyperdrive/blob/main/hyperdrive/packages/app-store/app-store/src/http_api.rs#L410) but the App Store backend is pretty complicated (the most complicated package core has worked on for sure), so please don't rabbithole too much here; I'm happy to assist.

This is lower priority than App Store public IMO

0x70b1a5:
[...]

PR #822, opened 2025-08-18
Cross app links

0x70b1a5:
Addresses #802, #805.

#811 isn't quite done yet, unless I misunderstand the reqs:
- ✅  With this PR, the groundwork exists to let any app open apps by sending a message to the parent iframe. 
- ❌  However, we lack a way to automagically invoke the message-send on a `hd://blah-address` link click.  The frontend that wishes to open the app must manually invoke the message to the parent frame.

It might be that this state of affairs is sufficient.  If not, however, I'd ask for @nick1udwig to discuss with me the intended end-state, eg "arbitrary link click panopticon intercepts `hd://` protocol `a[hrefs]`" or something ... it's a developer ergonomics q. 

## Testing 
1. open the appstore 
2. click the install button and verify that you're navigated to the app's dedicated page, and that the install action is automatically initiated.
3. If the install errors, the page should NOT blank out with a single message smack-dab in the middle anymore; instead, informative notifications should pop into the notif bay at top-right, and a toast with a preview of the message should also appear, in a dark or light mode corresponding to your browser's preference.
4. If the install is successful, verify that the Launch button appears on the app's store page. 
5. Click Launch. 
6. Verify the iframe opens in an app drawer as other apps opened from the homepage do. 
7. Verify that tapping the Omnibutton takes you to the Recent Apps view, displaying your newly-installed opened app therein.
8. Verify tapping the Omnibutton again closes the Recent view and returns you home.
9. (bugfix) Verify that tapping the Omnibutton from the Homepage opens Recents as expected, and doesn't flash it for a frame and then close it again.
10. Back in the Appstore, page through the main grid to an app you've already got installed with a Launch button that is visible in the card.
11. Verify that when you click the Launch button in the app card, you are NOT taken to the app's App Store page, but rather that the app opens in a frame as from home.
12. Omnibutton click -> verify Launched app is present in the Recents list, alongside the initially opened app (if present).
13. On mobile, verify swiping left or right closes the apps.


nick1udwig (changes requested):
## The good:

- `Launch` works great on non-localhost

## To fix before merge:

- [x] `Launch` does nothing on localhost: this is because /main:app-store:sys is opened in its own page, not an iframe within homepage. Ideally it would open the app in a new tab
- [x] `Install` doesn't work for me from the "apps view". It works fine from the app page itself, but when I click it from the "apps view" it reports a number of bad things (see attached). It downloads different files from the same publisher. I THINK what is happening is it is trying to download the same thing twice somehow

## Ideal thing we are aiming for:

- Render links like `hd://main:app-store-sys/foo` as clickable and do the correct thing when they are clicked. So an app need only write that link and it will Just Work. E.g. imagine you and I are chatting in a chat app and I send you a link of that form: we ideally want to render it as clickable
- How technically feasible is this?

## Install issues error messages:

https://gist.github.com/nick1udwig/1b4f8f0ea8b094dd9ca0b7a2d73c9ff8

0x70b1a5:
@nick1udwig this localhost thing is NUTS 🌰🌰🌰 ! Special code paths have a bad smell imo - or maybe it's just because these days, i'm sshing from laptop into home and cannot actually test it during the mornings - but can you think of a downside for home-rolling users if we redirected localhost to 127.0.0.1?

0x70b1a5:
@nick1udwig I'm seeing some strange things with installs - the Install intent works for Hypergrid, but not for Wifenote. I'll maybe pull you in for an assessment tomorrow morning if possible.

nick1udwig (changes requested):
Very close! Install button works great on localhost & non-localhost. Launch button works great on non-localhost. On localhost, it only works for me on wifenote (I think because the current version of wifenote publishes its frontend without auth :sweat_smile: ; this is not an intended behavior: wifenote should also not work); other apps say `secure subdomain can only serve its associated package`

I'm wondering if for localhost we can both:
1. open in a new tab
2. open without any subdomain

The problem here is with 2: http://app-store-sys.localhost:8080/fwd-ws:kibitz:nick.hypr/index.html is explicitly not allowed by http-server. 1 is just a QOL thing

nick1udwig (approved):
Beautiful, very nice work :rocket:

PR #829, opened 2025-08-25
Dw/hmap binding cacher

dolled-possum:
## Problem

Binding entries from the tokenregistry contract will need be used by other processes in the near future.

## Solution

While the work has not been made in process_lib to exploit the information from the tokenregistry contract, this work runs a cacher for the data, and ensures that the scripts associated with hypermap-cacher automatically affect the binding-cacher process as well.

## Testing

SIMULATION MODE ONLY: On super-debug verbosity, it's easy to see the impact of the start-providing, stop-providing, set-nodes, and reset-cache scripts now affecting both hypermap-cacher and binding-cacher processes.  The hypermap-cache and binding-cache data are now stored in their own respective directories in vfs/hypermap-cacher:sys, and the cached records can be directly viewed there.

## Notes

Because this work has incremented the version of the .wit for to add the binding-cacher interface to the hypermap-cacher world, the Cargo.toml references of the package now point to a specific commit in the process_lib repo (in the dw/hmap-binding-cacher branch, commit xxxxxxx) UPDATE: now xxxxxxx UPDATE: will just update in the thread below.  This branch also defines the contract address for the tokenregistry contract when deployed per the 02-DeployTokenRegistry.s.sol:DeployTokenRegistry script in protocol.


dolled-possum:
process_lib repo (in the dw/hmap-binding-cacher branch, commit e11d17cad18618117f1ef83520e71d46275f0504)

dolled-possum:
New plans after First Contact With the Enemy:

1. Restore the original cache directory name for smoother upgradability. (MUST remember to always keep this in mind.)
2. Add support for a `--use-no-defaults` flag:
2.1. Must be used in conjunction with `--rpc||--rpc-config` and `--cache-source||--cache-source-config` (`--cache-source` doesn't exist yet, but maybe add for an easy parallel entry method like `--rpc` is).
2.2. Then actually add the functionality of suppressing the defaults if used.  On the flip side, the mere presence of `--rpc||--rpc-config` or `--cache-source||--cache-source-config` does NOT mean to suppress defaults, just add the extra ones to the front of the list.
3. Possibly add in the router overridability functionality as well, or maybe leave that to another PR.

dolled-possum:
This is way outdated due to extensive hypermap-cacher work in the interim.  I'll keep this PR around while I build up a new branch, and delete it when the new one is ready.

dolled-possum:
Obsolete, replaced with #869

PR #833, opened 2025-08-27
fix #831

### LINKED ISSUE #831: bug: register-ui /login should redirect to / when there is no node id
Image

0x70b1a5:
## Problem

When there was no `hnsName`, the login page would display an error.

## Solution

Redirect to `/` in the event we do not have a name.


nick1udwig (approved):
Ty sir :saluting_face:

PR #835, opened 2025-08-27
fix #816

### LINKED ISSUE #816: bug: app-store: widget opens app store in a new tab, rather than in an iframe

0x70b1a5:
## Problem

Clicking into the appstore widget resulted in a newtab on non-localhost 

## Solution

When possible, use the window messaging protocol to open an iframe instead. Localhost should still use a newtab, similar to #822


nick1udwig (approved):
Works well, thank you :rocket:

PR #836, opened 2025-08-28
Dw/default sources

dolled-possum:
## Problem

Allow users to specify default RPC and node cache source overrides so that they aren't stuck using our defaults

## Solution

RPC already allows default overriding, now cache source does do.

Still requires doc, will edit this description and add it as I test and review.


hyperdrive/packages/hypermap-cacher/hypermap-cacher/src/lib.rs
```diff
@@ -146,7 +159,11 @@ impl State {
             is_cache_timer_live: false,
             drive_path: drive_path.to_string(),
             is_providing: false,
-            nodes: DEFAULT_NODES.iter().map(|s| s.to_string()).collect(),
+            nodes: load_cache_sources(std::env::var("CACHE_SOURCE_CONFIG_PATH").ok().as_deref())
```
nick1udwig, line 162: I'd be surprised if this works: I think we need to do some special things when starting a process for it to see env vars

nick1udwig:
Was this replaced by #858? If so can we close this?

dolled-possum:
Oh right, #858 replaced this one too!  Closing...

PR #840, opened 2025-09-03
app protocol links + semihelpful refresh

0x70b1a5:
## Problem

1. #811 
2. hitting Refresh on an open app closes the app

## Solution

1. implement #811
2. read the `#url-hash` if present to attempt to find an app to open on refresh

### #811
- [ ] open the Appstore Widget. Click a listing - they use `hw://` links now. 
- [ ] Verify the app listing click opens the appstore page of the respective app.
- Without closing the appstore, proceed immediately to the next set of tests.

### Reloads
- [ ] Hit Refresh, CTRL+R or F5. 
- [ ] Verify that the appstore reopens **to the main appstore page** on the page reload. (sub-page nav is something i'd like to add later on if needed - it's possible for the appstore we have but it may not be desired in the fully general case of all webapps.)

### DIFF
--- a/hyperdrive/packages/app-store/app-store/src/http_api.rs
+++ b/hyperdrive/packages/app-store/app-store/src/http_api.rs
@@ -174,6 +174,83 @@ fn make_widget() -> String {
     <h3>Top Apps</h3>
     <div id="latest-apps"></div>
     <script>
+    
+class HWProtocolWatcher {
+    constructor() {
+        this.isListening = false;
+        this.handleClick = this.handleClick.bind(this);
+    }
+    start() {
+        if (this.isListening) {
+            return;
+        }
+        document.addEventListener('click', this.handleClick, true);
+        console.log('[hw-protocol-watcher] initialized');
+        this.isListening = true;
+    }
+    stop() {
+        if (!this.isListening) {
+            return;
+        }
+        document.removeEventListener('click', this.handleClick, true);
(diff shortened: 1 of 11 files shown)


nick1udwig (approved):
Really nice work

I have a small nit for you to look at before merge:

- Click an app in widget
- App Store properly opens to that app's page
- Click to go to App Store home
- Click action button and go to home screen (leaving App Store open on the home page)
- Click on an app in widget
- App Store opens, but not to the app's page: instead just to the homepage

Is this intended or expected behavior? Is it easy to fix?

0x70b1a5:
@nick1udwig It should work now. There's a trade-off here where I believe the issue is:

- the homepage only has control over the iframe's top-level `src` attr
- the homepage does NOT have the ability to manipulate child SPA paths js-wise 
  - such a thing could be done, but it'd involve more iframe message passing and `hw://` library updates
- anyway, changing this `src` attr does indeed "navigate" the iframe (by causing it to **entirely reload**)
- this results in a slow, but adequate, completion of the desired action

I vote "make it exist first, make it good later" here, personally. If we need lightning fast intra-app switching via external link clicks, it might be better to put the work into it once there is a solid user story informed by the need.

nick1udwig:
@0x70b1a5 not going to hold up the PR on this, but there is still one edge case, see https://github.com/hyperware-ai/hyperdrive/issues/843

PR #858, opened 2025-09-16
Custom router set selection for indirect nodes, cache sources, and Base L2 Access URLs (Dw/router designation)

dolled-possum:
## Problem

Hyperdrive relies on some difficult- or impossible-to-override defaults for routers, Base L2 access points, and cache sources, many of which are our own infrastructure.  This is not tenable long-term.

## Solution

The CommitDotOsName, MintCustom, and ResetName pages now feature a Network Options section for overriding routers if desired, and the Login (register-served, not hyperdrive-served) and SetPassword pages feature a new Advanced Options for setting cache sources and Base L2 wss:// access points. 

## Testing

These features should all behave intuitively.  The front end will enforce that the inputs are sane, but they do not check to ensure that the nodes or URLs provided can actually provide the services requested.

## Docs Update

Docs must be updated to capture this new ability.

## Notes

The Base L2 access URLs are only added to the existing providers (there is a deduplication step, so adding an existing URL moves it to the top of the list), but currently only the terminal script "remove-provider" (or directly editing .eth_providers) will remove Base L2 access providers.

A future --no-defaults launch flag would be a good complementary addition to prevent accidental defaulting for those who do not want to or cannot use our infrastructure.


nick1udwig (approved):
two minor comments, then go ahead and merge. Really nice work :rocket: 

## tos

image

"By registering a name, you agree to the terms of service at [hyperwareware.xyz/tos](https://hyperwareware.xyz/tos)."

remove the tos

## indirect node specification prose

image

lets change the prose here

1. Network Options -> Advanced Network Options
2. "Register as indirect node with non-default routers."

dolled-possum:
Fixes made (ugh, that TOS one was an embarrassing miss) and pushed to this branch, but holding off on merging and closing because the latest "development" commits no longer compile successfully for me (neither does a clean clone of the "develop" branch HEAD).

dolled-possum:
All set.  Merging.

PR #864, opened 2025-09-30
fixed hyperwallet gas caps for bundler's gas response

Hallmane:
## Problem
Hyperwallet had a random gas cap of 55,000 gwei on bundler sends even after getting correct bundler gas estimates.

## Solution
I removed the cap. 

## Misc
Gas estimation is a module that could be centralized a bit more inside hyperwallet.


nick1udwig (approved):
Ty for fix :saluting_face:

PR #866, opened 2025-10-02
added upgrade hyper account view

pgreen88:
## Problem

no way to upgrade an existing account

## Solution

added new view


nick1udwig:
We need more information on this page about what is going on. What does it mean to "Upgrade Hyper Account"? What is the purpose of this page? Describe it on the page. If you want, you can either add commit to do so in code or describe it here and I can add prose.

pgreen88:
Ok, will add more info

PR #867, opened 2025-10-03
fixed unexpected submit bug on clicking back button

pgreen88:
Bug fix


nick1udwig (approved):
Very nice, thanks for the fix! :rocket:

PR #869, opened 2025-10-09
Dw/binding cacher

dolled-possum:
## Problem

Binding entries from the tokenregistry contract will need be used by other processes in the near future.

## Solution

While the work has not been made in process_lib to exploit the information from the tokenregistry contract, this work runs a cacher for the data, and ensures that the scripts associated with hypermap-cacher automatically affect the binding-cacher process as well.

## Testing

On super-debug verbosity, it's easy to see the impact of the start-providing, stop-providing, set-nodes, and reset-cache scripts now affecting both hypermap-cacher and binding-cacher processes.  The hypermap-cache and binding-cache data are now stored in their own respective directories in vfs/hypermap-cacher:sys, and the cached records can be directly viewed there.

## Notes

Because this work has incremented the version of the .wit for to add the binding-cacher interface to the hypermap-cacher world, the Cargo.toml references of the package now point to a specific commit in the process_lib repo (in the dw/hmap-binding-cacher branch, commit [ac36caf](https://github.com/hyperware-ai/process_lib/commit/ac36cafd98266548b430ce95fde71c36c27779d7))


hyperdrive/packages/hypermap-cacher/binding-cacher/src/lib.rs
```diff
@@ -0,0 +1,1491 @@
...
+        },
+        Err(e) => {
+            info!("Failed to open cache_sources: {}", e);
+        }
+    }
+
+    let bind_config = http::server::HttpBindingConfig::default().authenticated(false);
+    let mut server = http::server::HttpServer::new(5);
+
+    let hypermap_provider = hypermap::Hypermap::default(60);
```
nick1udwig: why not use a [eth::Provider](https://github.com/hyperware-ai/process_lib/blob/764316b298f181e5560c2103296e8f6db69cb4d6/src/eth.rs#L313) instead of hypermap::Hypermap since we only ever use the provider anyways?

hyperdrive/packages/hypermap-cacher/binding-cacher/src/lib.rs
```diff
@@ -0,0 +1,1491 @@
...
+    // Bootstrap state from other nodes, then fallback to RPC
+    fn bootstrap_state(&mut self, hypermap: &hypermap::Hypermap) -> anyhow::Result<()> {
+        info!("Starting state bootstrap process...");
+
+        // Try to bootstrap from other nodes first
+        if let Ok(()) = self.try_bootstrap_from_nodes() {
+            info!("Successfully bootstrapped from other nodes");
+        }
+
+        self.try_bootstrap_from_rpc(hypermap)?;
```
nick1udwig: before we try to bootstrap from rpc we should wait until hypermap-cacher is done bootstrapping. This avoids case where both cachers are hitting rpc and rpc rate limits us

examples of processes doing this are [spider](https://github.com/hyperware-ai/hyperdrive/blob/c49d3e720fc8c6b099e0eaf6011fcff793f7d579/hyperdrive/packages/spider/spider/src/lib.rs#L105-L153) and [hns-indexer](https://github.com/hyperware-ai/hyperdrive/blob/c49d3e720fc8c6b099e0eaf6011fcff793f7d579/hyperdrive/packages/hns-indexer/hns-indexer/src/lib.rs#L190) (where hns-indexer uses process_lib, see [here](https://github.com/hyperware-ai/process_lib/blob/main/src/hypermap.rs#L849-L882) and [here](https://github.com/hyperware-ai/process_lib/blob/main/src/hypermap.rs#L961))

note that the "retry on error" is very important here because I expect we'll get `Timeout`s not actually `IsStarting`s if hypermap-cacher is in the middle of its bootstrap (since it will not be handling incoming messages)

it may be worth taking the pattern from spider and putting it into a process_lib function alongside bootstrap, but processes that want to delay until hypermap-cacher is ready but don't care to bootstrap from it (like spider and binding-cacher)

nick1udwig (changes requested):
a couple comments, the biggest to delay rpc until hypermap-cacher is ready

hyperdrive/packages/hypermap-cacher/binding-cacher/src/lib.rs
dolled-possum: It would certainly be possible.  I'd have to reference the chain IDs and starting block numbers slightly differently, but that's doable.  Mostly, it's because hypermap-cacher does it this way as well and I was keeping parity.  Let me know if you'd like me to make the changes in both processes, but I don't think there's a good reason to only change one.

hyperdrive/packages/hypermap-cacher/binding-cacher/src/lib.rs
dolled-possum: I really liked the idea of making it a process_lib helper, and I liked even more the idea of making it a generic "wait for process readiness" function.  So that's been added, and the binding-cacher now uses it to wait for hypermap-cacher to be ready in the same way that spider was doing it in its own code.  Should I also update spider to use this helper now that it's available?

nick1udwig: Nice! Let's just make an issue for spider

hyperdrive/packages/hypermap-cacher/binding-cacher/src/lib.rs
nick1udwig: Agreed -- both should be provider, not hypermap

nick1udwig (approved):
replied to your comments -- otherwise lgtm, merge after the minor tweaks we've discussed

hyperdrive/packages/hypermap-cacher/binding-cacher/src/lib.rs
dolled-possum: https://github.com/hyperware-ai/todo/issues/1

PR #874, opened 2025-10-14
added support for building core packages in parallel with ryon

Gohlub:
## Problem

I doing some work on core and got tired of waiting to rebuild after each fix (there were ways to make it easier but I just stuck with it).

## Solution

Solution was elegant: substitute the iterator method call from `into_iter' to 'into_par_iter' (from the rayon crate), which handles the parallel execution under the hood. Tested it and with no special flags (you can specify the number of cores)  got a 2x speedup in the build-package script time (from 6 minutes to sub 3 minutes on a re-build).

## Testing
Could be made default but for the PR I left is as an optional flag.
```
cargo run -p build-packages -- --parallel
```

## Docs Update

Reflected in README

## Notes

This was downstream of me just asking an LLM client to 'make it faster'. I wonder if we could apply the same principle to other places in our codebase that have long build times,


nick1udwig (approved):
Really nice, thanks for the PR :zap:

PR #900, opened 2025-12-03
app-store: add bindings

nick1udwig:
## Problem

How to avoid App Store showing out-of-date or low-quality apps?

## Solution

Bindings. So implement it!

## Testing

TODO

## Docs Update

TODO

## Notes

None


nick1udwig:
no binding power sorting:

image

yes binding power sorting (I bound to HYPR DAO):

image

